Issue #65, opened 2014-05-20
igb_main.c fails to compile against a kernel newer than v3.14


jinksong:
compare_ether_addr() is removed since kernel v3.14. ether_addr_equal() is used instead, while it has an opposite polarity in return value.
https://lkml.org/lkml/2013/12/5/595

Build error is pasted here:
[   59s] /home/abuild/rpmbuild/BUILD/openavb-20130814/kmod/igb/igb_main.c:2075:5: warning: initialization from incompatible pointer type [enabled by default]
[   59s]      .ndo_select_queue       = igb_select_queue,
[   59s]      ^
[   59s] /home/abuild/rpmbuild/BUILD/openavb-20130814/kmod/igb/igb_main.c:2075:5: warning: (near initialization for 'igb_netdev_ops.ndo_select_queue') [enabled by default]
[   60s] /home/abuild/rpmbuild/BUILD/openavb-20130814/kmod/igb/igb_main.c: In function 'igb_del_mac_filter':
[   60s] /home/abuild/rpmbuild/BUILD/openavb-20130814/kmod/igb/igb_main.c:9051:3: error: implicit declaration of function 'compare_ether_addr' [-Werror=implicit-function-declaration]
[   60s]    if (!compare_ether_addr(addr, adapter->mac_table[i].addr) &&
[   60s]    ^

Please make the kernel module work with the latest upstream kernel.

andrew-elder:
Fixed in pull request number 67.

PR #67, opened 2014-05-28
kmod: Make kmod compatible with kernel 3.14 or later

jinksong:
Since linux kernel 3.14, compare_ether_addr() is deprecated and
ether_addr_equal() is used instead.
And function definition of igb_select_queue is also modified.

Signed-off-by: Jin Song jin.kyu.song@linux.intel.com


ahogen:
This was reported as issue #65. Can #65 be closed now, then?

andrew-elder:
Thanks - #65 is now closed.

PR #142, opened 2014-11-07
daemons: maap: Makefile fix build error

prabhakarlad:
The rule to build maap daemon tries to copy the executable
in the build directory, since there wasnt any build directory
the rule was failing to build the maap daemon.
This patch fixes it by creating a build directory, while building
the maap daemon, and removes it off on calling the clean rule.

Signed-off-by: Lad, Prabhakar <prabhakar.csengg@gmail.com>


pinealservo:
I'm closing this, as there's been no action on it in quite a while and it was most likely fixed another way.

Issue #169, opened 2015-01-02
MRP regression: listener does not see talker


jbmaillet:
A regression I introduced myself I'm affraid, by a typo in 725f9eb2.
Pull request for fix, #170

davidcemin:
It looks like this issue was already merged. Can we close it?

PR #170, opened 2015-01-02
bugfix issue 169: MRP regression: listener does not see talker

jbmaillet:
Fatal typo: '=' instead of '=='.


jbmaillet:
Fix for issue #169

PR #171, opened 2015-01-05
Add support for Linux "generic" cross timestamp support

christopher-s-hall:


andrew-elder:
Chris, did you see the build failed?

christopher-s-hall:
I did.  Which distribution is being used?  The ‘broken’ #defines should be in <linux/ptp_clock.h> which is included.  The compile works fine for me on 14.04.

andrew-elder:
See
<https://travis-ci.org/AVnu/Open-AVB/builds/45966497>

On 1/5/2015 12:16 PM, Christopher Hall wrote:
> I did. Which distribution is being used? The ‘broken’ #defines should be 
> in <linux/ptp_clock.h> which is included. The compile works fine for me 
> on 14.04.
>
> From: andrew-elder [mailto:notifications@github.com]
> Sent: Monday, January 05, 2015 8:44 AM
> To: AVnu/Open-AVB
> Cc: Hall, Christopher S
> Subject: Re: [Open-AVB] Add support for Linux "generic" cross timestamp 
> support (#171)
>
>
> Chris, did you see the build failed?
>
> —
> Reply to this email directly or view it on 
> GitHub<https://github.com/AVnu/Open-AVB/pull/171#issuecomment-68735554>.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/AVnu/Open-AVB/pull/171#issuecomment-68740771>.
>


-- 
Andrew Elder
AudioScience, Inc.
+1-585-271-8870
www.audioscience.com

christopher-s-hall:
It appears that Ubuntu 14.04 is being used.  How do I get more information about the failure?  Again, this builds fine on my local system.

andrew-elder:
I see the line
 export BUILD_KERNEL=3.13.0-36-generic
in the log. Doesn't that mean 3.13.0-36 kernel is used for the build?

intel-ethernet:
I think 3.13.0-36 represents the patch level of a distribution-specific release e.g. Ubuntu backports patches from 3.17 into its 3.13 release … you need to define the distribution as well to fully qualify the release.

ekm

pinealservo:
On Mon, Jan 5, 2015 at 10:16 AM, Christopher Hall <notifications@github.com>
wrote:

> I did. Which distribution is being used? The ‘broken’ #defines should be
> in <linux/ptp_clock.h> which is included. The compile works fine for me on
> 14.04.
>

If you look at the .travis.yaml file, you'll see there's already one
workaround in there to deal with the old Ubuntu image Travis uses (12.04
when the build was set up... perhaps they updated to the new LTS now,
though?) where a header file is copied from a newer kernel install into the
/usr/include/linux directory. By default, /usr/include/linux on Ubuntu is
based on the kernel that the release shipped with rather than the currently
installed one, so builds that require newer kernel headers will fail
without some sort of workaround.

If you add a similar line to the .travis.yaml file to move the ptp_clock.h
header, you should be able to get the build to work.

      --Levi

christopher-s-hall:
After some investigation (using the version number spat out by gcc) I determined that the Ubuntu distro version is 12.04.

I can confirm that the build does fail on this.  Is there some reason that we are using such an old (nearly 3 years old) version?  Can we upgrade to, say, 14.04?  This is a “LTS” version if that is a concern.

Chris

pinealservo:
On Mon, Jan 5, 2015 at 10:45 AM, Christopher Hall <notifications@github.com>
wrote:

> It appears that Ubuntu 14.04 is being used. How do I get more information
> about the failure? Again, this builds fine on my local system.
>
>
Andrew gave the following link: <
https://travis-ci.org/AVnu/Open-AVB/builds/45966497>

If you click on the little arrow next to the yellowish "Build system
information" line, it will expand all the build system info.

You'll find it's still using Ubuntu 12.04, not 14.04.

      --Levi

pinealservo:
On Mon, Jan 5, 2015 at 11:25 AM, Christopher Hall <notifications@github.com>
wrote:

> After some investigation (using the version number spat out by gcc) I
> determined that the Ubuntu distro version is 12.04.
>
> I can confirm that the build does fail on this. Is there some reason that
> we are using such an old (nearly 3 years old) version? Can we upgrade to,
> say, 14.04? This is a “LTS” version if that is a concern.
>
> Chris
>

Because that's what's installed on their build server images, which we
can't change.  It's old, but only in the middle of its support lifetime, so
it gets kernel refreshes.  The kernel API is the only real build problem,
and my previous email describes how to get around that.

      --Levi

PR #175, opened 2015-01-20
Update jackd_talker.c

sumeet-jain-engineer:
Without the above change, igb_xmit fails as no tx_queues are allocated/mapped.

### DIFF
--- a/examples/jackd-talker/jackd_talker.c
+++ b/examples/jackd-talker/jackd_talker.c
@@ -766,8 +766,13 @@ int pci_connect()
 		}
 		printf("attaching to %s\n", devpath);
 		err = igb_attach(devpath, &igb_dev);
-		if (err || igb_attach_tx( &igb_dev )) {
+		if (err) {
 			printf("attach failed! (%s)\n", strerror(errno));
+			continue;
+		}
+		err = igb_attach_tx(devpath, &igb_dev);
+		if (err) {
+			printf("igb_attach_tx failed! (%s)\n", strerror(errno));
 			continue;
 		}
 		goto out;
--- a/examples/jackd-talker/jackd_talker.c
+++ b/examples/jackd-talker/jackd_talker.c
@@ -766,7 +766,7 @@ int pci_connect()
 		}
 		printf("attaching to %s\n", devpath);
 		err = igb_attach(devpath, &igb_dev);
-		if (err) {
+		if (err || igb_attach_tx( &igb_dev )) {
 			printf("attach failed! (%s)\n", strerror(errno));
 			continue;
 		}


examples/jackd-talker/jackd_talker.c
```diff
@@ -766,7 +766,7 @@ int pci_connect()
 		}
 		printf("attaching to %s\n", devpath);
 		err = igb_attach(devpath, &igb_dev);
-		if (err) {
+		if (err || igb_attach_tx( &igb_dev )) {
```
andrew-elder: This logic looks confusing. Why wouldn't the error from igb_attach_tx() be handled more clearly?

grandcat: For me, it's also confusing. However, I noticed that it is done the same way for other examples like the `simple_talker.c` (line 379):

```c
		if ( err || igb_attach_tx( igb_dev )) {
			printf("attach failed! (%s)\n", strerror(errno));
			continue;
		}
```

sumeet-jain-engineer:
@andrew-elder I agree that the error returned by igb_attach_tx should have been handled appropriately. But as the call to this function is missing in 'jackd_talker.c' application, 'igb_xmit' returns from:-
txr = &adapter->tx_rings[queue_index];
	if( !txr ) return EINVAL;

andrew-elder:
@sumeet-jain-engineer I do understand that the patch is required. I'd just like the err handling to be a little cleaner. When strerror(errno) will you be able to tell whether igb_attach() or igb_attach_tx() failed?

The below is cleaner in my opinion. What do you think?

if ( err ) {
            printf("igb_attach failed! (%s)\n", strerror(errno));
            continue;
}
err = igb_attach_tx(devpath, igb_dev);
 if ( err ) {
            printf("igb_attach_tx failed! (%s)\n", strerror(errno));
            continue;
  }

sumeet-jain-engineer:
@andrew-elder 
It looks better now !

andrew-elder:
@sumeet-jain-engineer can you update your pull request?

sumeet-jain-engineer:
@andrew-elder  Sure !

grandcat:
Is it not like this that `igb_attach_tx` has to be attached if and only if `igb_attach` was not successful?
This is the logic how it was before in the simple_talker example.

sumeet-jain-engineer:
@grandcat 
I feel the modification that andrew suggested can be justified with the logic that there is no point in calling igb_attach_tx if igb_attach fails.

pinealservo:
I like the new code as this currently sits, and it seems preferable to the original condition to me as well.  But for future reference....

The original code pattern was familiar to me as a way to de-nest a chained conditional. These crop up frequently when you follow certain coding guidelines, such as avoiding multiple exit paths from a control structure. Although it's not a super-clear construct and not my favorite way to deal with those situations in C, it's got definite advantages over deeply nested conditionals that are really just a linear sequence of tests if you're not allowed to use the other facilities that C offers to deal with that situation.

That justification for it doesn't really apply here, since `continue` and `goto` are already in play, so I fully support the rewrite. But it's a handy construct to be familiar with if you're ever stuck with a very restrictive style guide and you have a long series of conditional tests to make.

PR #176, opened 2015-01-20
Improved documentation for talker/listener example

grandcat:
For the talker/listener example, I improved the documentation.
Now, it should be possible to setup two AVB instances by plain copy and paste.

I think it is helpful for newcomers.


andrew-elder:
This looks good to me. Anyone else?

davidcemin:
Looks good, its more clear now.

PR #177, opened 2015-01-21
Fixed AVnu certification test failure 6.1a and 6.3a

christopher-s-hall:
Rejects Announce message with sourcePortIdentity same as my own

AVnu certification test 6.3a - PASS
Rejects Path Trace TLV with ClockIdentity same as mine
Implemented PathTraceTLV Parsing
Implemented PathTraceTLV Append/Search Operations
Added debug (print method) to ClockIdentity class

Corrected misc output format (printf), indentation issues

### DIFF
--- a/daemons/gptp/common/avbts_message.hpp
+++ b/daemons/gptp/common/avbts_message.hpp
@@ -269,6 +269,9 @@ class PathTraceTLV {
 			identityList.end();
 	}
 	int length() {
+		// Total length of TLV is length of type field (UINT16) + length of 'length' field (UINT16) + length of
+		// identities (each PTP_CLOCK_IDENTITY_LENGTH) in the path
+		// See IEEE 802.1AS doc table 10-8 for details
 		return 2*sizeof(uint16_t) + PTP_CLOCK_IDENTITY_LENGTH*identityList.size();
 	}
 };
--- a/daemons/gptp/common/avbts_message.hpp
+++ b/daemons/gptp/common/avbts_message.hpp
@@ -226,6 +226,9 @@ protected:
 
 #pragma pack(push,1)
 
+// Path Trace TLV
+// See IEEE 802.1AS doc table 10-8 for details
+
 #define PATH_TRACE_TLV_TYPE 0x8
 
 class PathTraceTLV {
@@ -252,9 +255,10 @@ class PathTraceTLV {
 	}
 	void toByteString(uint8_t * byte_str) {
 		IdentityList::iterator iter;
-		*((uint16_t *)byte_str) = tlvType;
+		*((uint16_t *)byte_str) = tlvType;  // tlvType already in network byte order
 		byte_str += sizeof(tlvType);
-		*((uint16_t *)byte_str) = PLAT_htons(identityList.size()*PTP_CLOCK_IDENTITY_LENGTH);
+		*((uint16_t *)byte_str) =
+			PLAT_htons(identityList.size()*PTP_CLOCK_IDENTITY_LENGTH);
 		byte_str += sizeof(uint16_t);
 		for
 			(iter = identityList.begin();
@@ -269,9 +273,9 @@ class PathTraceTLV {
 			identityList.end();
 	}
 	int length() {
-		// Total length of TLV is length of type field (UINT16) + length of 'length' field (UINT16) + length of
+		// Total length of TLV is length of type field (UINT16) + length of 'length'
+		// field (UINT16) + length of
 		// identities (each PTP_CLOCK_IDENTITY_LENGTH) in the path
-		// See IEEE 802.1AS doc table 10-8 for details
 		return 2*sizeof(uint16_t) + PTP_CLOCK_IDENTITY_LENGTH*identityList.size();
 	}
 };
--- a/daemons/gptp/common/avbts_message.hpp
+++ b/daemons/gptp/common/avbts_message.hpp
@@ -38,6 +38,9 @@
 #include <avbts_osnet.hpp>
 #include <ieee1588.hpp>
 
+#include <list>
+#include <algorithm>
+
 #define PTP_CODE_STRING_LENGTH 4
 #define PTP_SUBDOMAIN_NAME_LENGTH 16
 #define PTP_FLAGS_LENGTH 2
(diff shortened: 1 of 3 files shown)


daemons/gptp/common/avbts_message.hpp
```diff
@@ -223,21 +226,50 @@ class PTPMessageCommon {
...
+			byte_str += PTP_CLOCK_IDENTITY_LENGTH;
+		}
+	}
+	bool has(ClockIdentity *id) {
+		return std::find
+			(identityList.begin(), identityList.end(), *id) !=
+			identityList.end();
+	}
+	int length() {
+		return 2*sizeof(uint16_t) + PTP_CLOCK_IDENTITY_LENGTH*identityList.size();
```
andrew-elder: comment on magic 2* would be good. I'm guessing sizeof(tlvType) + sizeof(length)?

daemons/gptp/common/avbts_message.hpp
```diff
@@ -223,21 +226,50 @@ class PTPMessageCommon {
...
+			buffer += PTP_CLOCK_IDENTITY_LENGTH;
+		}
+	}
+	void appendClockIdentity(ClockIdentity * id) {
+		identityList.push_back(*id);
 	}
 	void toByteString(uint8_t * byte_str) {
-		memcpy(byte_str, this, sizeof(*this));
+		IdentityList::iterator iter;
+		*((uint16_t *)byte_str) = tlvType;
```
andrew-elder: It is a little confusing to apply htons() to some variables and not others

daemons/gptp/common/avbts_message.hpp
```diff
@@ -223,21 +226,50 @@ class PTPMessageCommon {
...
+	PathTraceTLV() {
+		tlvType = PLAT_htons(PATH_TRACE_TLV_TYPE);
+	}
+	void parseClockIdentity(uint8_t *buffer) {
+		int length = PLAT_ntohs(*((uint16_t *)buffer))/PTP_CLOCK_IDENTITY_LENGTH;
+		buffer += sizeof(uint16_t);
+		for(; length > 0; --length) {
+			ClockIdentity add;
+			add.set(buffer);
+			identityList.push_back(add);
```
andrew-elder: I assume list "delete" is elsewhere?

daemons/gptp/common/avbts_message.hpp
```diff
@@ -223,21 +226,50 @@ class PTPMessageCommon {
...
-		memcpy(byte_str, this, sizeof(*this));
+		IdentityList::iterator iter;
+		*((uint16_t *)byte_str) = tlvType;
+		byte_str += sizeof(tlvType);
+		*((uint16_t *)byte_str) = PLAT_htons(identityList.size()*PTP_CLOCK_IDENTITY_LENGTH);
+		byte_str += sizeof(uint16_t);
+		for
+			(iter = identityList.begin();
+			 iter != identityList.end(); ++iter) {
+			iter->getIdentityString(byte_str);
```
andrew-elder: getIdentityString() is really writeIdentityToString()?

daemons/gptp/common/avbts_message.hpp
christopher-s-hall: You would only need a 'delete' if there is a 'new'.  The list is not dynamic and cleans itself up when delete is called on the message.

daemons/gptp/common/avbts_message.hpp
christopher-s-hall: I'll add a comment

daemons/gptp/common/avbts_message.hpp
christopher-s-hall: What is your suggestion?

daemons/gptp/common/avbts_message.hpp
christopher-s-hall: I'm deliberating trying to make as few changes as possible to get the tests to pass.  I would consider this to be cosmetic.  What's clear to you may not be clear to others, and so, on.

andrew-elder: Fair enough. I thought it was new code. Could we add a comment saying what is going here? I'm not sure I do understand, hence mycomment.

daemons/gptp/common/avbts_message.hpp
andrew-elder: Add a comment /* tlvType already in network byte order */. But that does bring up question for future documentation and/or resolution. What is the gPTP policy for storing variables? Are some random variables stored in network order? Code I have worked with stores all variables in machine order until such time as a packet is constructed. Only the packet under construction contains network ordered data.

christopher-s-hall: Good idea on the comment.  I don't have an answer for the larger question.  Mostly, I've tried to minimize the number of byte swaps.  In some places, I have appended an _no (network order) or _ho (host order to variables indicating what order they are stored in?  Would that help here?

pinealservo:
This is just a general comment/suggestion for future reference, rather than something you should put into this commit:

I like to encapsulate the formatting of data into byte buffers/byte streams into a small library of functions that handle the writing of data, byte-order conversions (if necessary), and buffer position pointer management.  This tends to make the functions that read/write various program data into the buffers much clearer and less error-prone, especially when you get into interactions of bit fields, byte order, and data field alignment. Casting arbitrary byte buffer positions to/from larger types can cause misaligned access exceptions on some architectures. E.g., from the ARM compiler support FAQ: "The compiler will generate code on the assumption that a pointer is correctly aligned." Most v7 ARM processors you're likely to run this on do support unaligned access, but notably Cortex-M processors still do not.

daemons/gptp/common/avbts_message.hpp
pinealservo: I also have a strong preference for consistency of byte order in memory rather than trying to optimize the number of swaps performed.  I would swap to native byte order when reading into a native data structure and swap to network byte order when writing back out to a network buffer.  The only exception I'd make would be for addressing real, measured performance problems.

I'd probably address performance issues by maintaining some templated buffers that just need a few fields filled in before sending; in this case, the network byte order storage is fine because it's in a network-bound buffer already.

andrew-elder:
MRP could benefit from pdu buffer read/write functions as well. I added as a wish to TWiki page.

daemons/gptp/common/avbts_message.hpp
andrew-elder: While I agree Levi, I'm ok with a comment for the short term.

PR #180, opened 2015-01-28
Feature MSRP pruning

andrew-elder:
Add support for optionally pruning TalkerAdvertise and TalkerFailed attributes from those actively maintained by mrpd.


andrew-elder:
Compiler failure - oops.

andrew-elder:
All, this pull request is now ready for review. Comments welcomed! Thanks.

daemons/mrpd/msrp.c
```diff
@@ -3222,6 +3277,20 @@ int msrp_dumptable(struct sockaddr_in *client)
...
 
+static int msrp_cmd_parse_stream_id(char *buf, int buflen,
+                                    uint8_t *stream_id,
+				    int *err_index )
+{
+	struct parse_param specs[] = {
+		{"S" PARSE_ASSIGN, parse_c64, stream_id },
+		{0, parse_null, 0 }
+	};
+	if( buflen < 16+4 )
```
pinealservo: Might be good to explain the magic numbers here and below.

daemons/mrpd/msrp.c
```diff
@@ -3625,8 +3790,18 @@ int msrp_init(int msrp_enable)
 	if (NULL == MSRP_db)
 		goto abort_socket;
 
+	if( max_interesting_stream_ids < 8 ) {
+		/* Allow minimum 8 interesting talker stream ids and 8 interesting listener stream ids */
```
pinealservo: We're only really tracking talker ids, right?

daemons/mrpd/msrp.c
andrew-elder: Agreed. Will fix.

daemons/mrpd/msrp.c
andrew-elder: Yes, talkerIDs are what the max_interesting applies to. So if you are suggesting I rename the  max_interesting_stream_ids variable, I agree.

pinealservo: I was actually referring to the comment, which suggests that there are 16 total stream_ids, 8 talker and 8 listener.

pinealservo:
I read through all the code, and it looks good to me. I like the tests, though I don't have the time to dig too deeply into how they're working. It looks like they provide reasonable coverage of the changes, though.

davidcemin:
I did read through all the code as well and all looks good.

pinealservo:
OK, looks like all the open comments were resolved.

Issue #184, opened 2015-02-03
Simple_talker (gptpinit): pointer by value instead of by reference for global scope


grandcat:
Hi all,

in the latest open-avb-next branch, the simple_talker/listener demo doesn't seem to work anymore.
After the client is detected and the transfer should start, it segfaults.

By debugging, I narrowed it down to the the function `gptpinit` in `simple_talker.c`.

```c
*igb_shm_fd = shm_open(SHM_NAME, O_RDWR, 0);
	if (*igb_shm_fd == -1) {
		perror("shm_open()");
		return -1;
	}
	igb_mmap = (char *)mmap(NULL, SHM_SIZE, PROT_READ | PROT_WRITE,
				MAP_SHARED, *igb_shm_fd, 0);
```
`igb_shm_fd` seems to be set with a valid value (e.x. 5), but `igb_mmap` will contain 0x0 after the call. This doesn't seem to be valid. There is no check therefore.

As a result, in `gptpscaling` which is the consecutive function to access this mapped memory area, it will segfault on trying to invoke `pthread_mutex_lock((pthread_mutex_t *) igb_mmap);`.

Is this a known problem?

grandcat:
Ok, I narrowed down one issue which is responsible for this problem. I will create a patch.

However, simple_talker --> simple_listener (without switch in my setup) won't transfer any data packages. But I assume this is another problem.

grandcat:
Is merged. Can be closed.

PR #185, opened 2015-02-04
Fixed segmentation fault: pointer adress not passed by reference

grandcat:
There is a critical segmentation fault which prevents that `simple_talker` and `jackd_talker` are working anymore.

`gptpinit()`: igb_mmap pointer is passed by value though its address is changed and should be updated in a global sense (not only within the function's scope).
Beside this problem, various minor warnings are fixed.

The corresponding issue is: #184 (though it's description is not adapted yet)

### DIFF
--- a/examples/simple_talker/simple_talker.c
+++ b/examples/simple_talker/simple_talker.c
@@ -32,6 +32,7 @@
 ******************************************************************************/
 
 #include <errno.h>
+#include <inttypes.h>
 #include <fcntl.h>
 #include <math.h>
 #include <signal.h>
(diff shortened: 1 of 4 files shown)


andrew-elder:
These changes look good to me. Anyone else?

examples/simple_talker/simple_talker.c
```diff
@@ -292,19 +293,21 @@ int gptpdeinit(int *igb_shm_fd, char *igb_mmap)
...
 {
 	if (NULL == td)
 		return -1;
 
-	pthread_mutex_lock((pthread_mutex_t *) igb_mmap);
-	memcpy(td, igb_mmap + sizeof(pthread_mutex_t), sizeof(*td));
-	pthread_mutex_unlock((pthread_mutex_t *) igb_mmap);
+	(void) pthread_mutex_lock((pthread_mutex_t *) igb_mmap);
+	{
+		memcpy(td, igb_mmap + sizeof(pthread_mutex_t), sizeof(*td));
```
pinealservo: I don't have a problem with this, but I'm curious whether it makes a functional difference in the code that's generated or if it was just done to highlight the mutex-protected function.

pinealservo:
Looks good to me. Thanks for the fix, @grandcat.

examples/simple_talker/simple_talker.c
grandcat: Hey,

as you noticed, this was simply done to highlight this critical section. It
is a common pattern to see its start and end.
Am 04.02.2015 18:18 schrieb "Levi Pearson" <notifications@github.com>:

PR #187, opened 2015-02-20
Markdown translation of Chris's gPTP design document

pinealservo:
It seems to have carried over pretty well.  I only had to fix one section of markup.

Here's the link to the file in my fork of Open-AVB so you can see how it renders: https://github.com/pinealservo/Open-AVB/blob/open-avb-next/documents/design/AVnu_gPTP_Code.md

### DIFF
--- a/documents/design/AVnu_gPTP_Code.asciidoc
+++ b/documents/design/AVnu_gPTP_Code.asciidoc
@@ -1,0 +1,330 @@
+= AVnu gPTP Code
+:toc:
+:toc-placement!:
+
+toc::[]
+
+[[introduction]]
+== Introduction
+
+[[conventions]]
+=== Conventions
+
+Text in red denotes information that is correct at the time the document
+of writing, but is likely to change
+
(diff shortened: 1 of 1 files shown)


andrew-elder:
Looks good to me. Is there and easy way to add a table of contents to the top to aid navigation?

pinealservo:
I made a version in asciidoc format instead of markdown to get a ToC; github's asciidoc renderer is a lot more powerful, and asciidoc is arguably better for technical writing anyway: https://github.com/pinealservo/Open-AVB/blob/open-avb-next/documents/design/AVnu_gPTP_Code.asciidoc

pinealservo:
If you're okay with one or the other, let me know and I'll clean up the history a bit and remove the other one.

andrew-elder:
I'm 100% ok with the asciidoc version. Thanks!

pinealservo:
OK, I cleaned up the history a bit and left only the asciidoc version.  See http://asciidoctor.org/ for the documentation on the asciidoc library that github uses; it's a pretty nice tool.

PR #196, opened 2015-03-28
Update avb.c

danjurious:
igb_attach_tx() call missing in common/avb.c

Signed-off-by: CK <danjurious@users.noreply.github.com>

### DIFF
--- a/examples/common/avb.c
+++ b/examples/common/avb.c
@@ -76,6 +76,13 @@ int pci_connect(device_t * igb_dev)
 			printf("attach failed! (%s)\n", strerror(errno));
 			continue;
 		}
+
+		err = igb_attach_tx(igb_dev);
+		if (err) {
+			printf("attach_tx failed! (%s)\n", strerror(errno));
+			continue;
+		}
+
 		goto out;
 	}
 


andrew-elder:
Anyone qualified to review this one?

pinealservo:
I object to it on code clarity grounds. It's bad enough using C's loose interpretation of non-zero error values as 'true' in conditionals, but that's at least an established idiom. Chaining two possibly-erroring calls via a short-circuiting boolean expression is clever and avoids repeating the error value checking, but the combination of interpreting error values as 'true' and putting a side-effecting statement in the right-hand side of a boolean conditional expression strikes me as going a bit too far.

examples/common/avb.c
```diff
@@ -72,10 +72,17 @@ int pci_connect(device_t * igb_dev)
 
 		printf("attaching to %s\n", devpath);
 		err = igb_attach(devpath, igb_dev);
-		if (err) {
+		if (err || igb_attach_tx(igb_dev)) {
```
andrew-elder: Is the igb_attach_tx() still required here?

danjurious: Yes, without that igb_xmit fails. We should probably have all example applications use these calls from common/avb.c. I'm guessing the common directory was created for this specific reason. At the moment it's there in multiple places
- simple_talker.c
- jackd_talker.c
- common/avb.c

@pinealservo simple_talker.c does this _if (err || igb_attach_tx())_. May be that needs to be fixed too. I got the simple talker/listener applications working first and then moved on to this one.

andrew-elder: But isn't the call to igb_attach_tx() present at line 80 below?

pinealservo: It is; that's apparently *also* part of this patch set, though.  It should be the only one; @danjurious, if you remove this chunk and leave the one that adds in the clearer version at line 80, I think it'll be acceptable.

danjurious: Oops! I realized I got my rebase wrong last time. Uploaded a new patch, hopefully it's all fine this time. Sorry!

PR #212, opened 2015-05-11
gPTP: add method to retrieve asCapable from IEEE1588Port class.

andrew-elder:
simple addition of a single method


andrew-elder:
I'd like to accept this pull request and then loop around to the topic of IPC changes after this pull request has been processed. Does anyone see any issues with that?

PR #221, opened 2015-05-28
Gptp doc task

davidcemin:


andrew-elder:
Mostly all looks fine. It does seem strange to have "Makefile" listed in gptp .gitignore. If I can make a suggestion, I think we should create a "build" subdir of gptp then go
cd build
cmake ..
and then we can just have "build" in our base .gitignore and do not worry about listing any files that are created when cmake runs. The whole idea of cmake is to support building from an out-of-tree location, so I think we should make use of that way of operating.

Anyone else have comments?

davidcemin:
Andrew,

sounds better. Should we integrate that pull request first ? Or i can just
fix that on the same branch and we integrate it on the same pull request...

let me know.

David

On Fri, May 29, 2015 at 9:04 AM, andrew-elder <notifications@github.com>
wrote:

> Mostly all looks fine. It does seem strange to have "Makefile" listed in
> gptp .gitignore. If I can make a suggestion, I think we should create a
> "build" subdir of gptp then go
> cd build
> cmake ..
> and then we can just have "build" in our base .gitignore and do not worry
> about listing any files that are created when cmake runs. The whole idea of
> cmake is to support building from an out-of-tree location, so I think we
> should make use of that way of operating.
>
> Anyone else have comments?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/AVnu/Open-AVB/pull/221#issuecomment-106797989>.
>



-- 
David Cemin
+1 343-777-1735

andrew-elder:
ChrisH reported via email that he is ok with this pull request!

PR #234, opened 2015-06-04
travis update to support gcc4.8

andrew-elder:


andrew-elder:
Still working on this. Please don't merge it yet....

andrew-elder:
Ok, this is reviewable now. I fixed the travis gcc version thing. I think it is pullable, but there are still more issues to be worked out.

pinealservo:
You may need a flag to tell gcc to build with the C99 or C11 standards; try with -std=c99 or -std=gnu99 (or c11/gnu11).  GCC 5 moved to gnu11 by default, but 4.8 is still at gnu89.  I believe clang, which is the default compiler on a Mac, builds with C11 by default, which is probably why it didn't bite Jeff.

andrew-elder:
Maybe I should just update to gcc 5.1?

pinealservo:
The gcc 5.1 release is still really new, and is not likely to be available by default on your average distro (especially embedded toolchains) for a while.  I'd suggest just adding the -std flags to the avdecc libs, since the c99/c11 support is just fine in 4.8, it's just not the default mode.

pinealservo:
Whoops, I didn't realize I had the AVnu repository branch as my upstream for the feature-avdecc branch. I meant to push it to my fork to test the travis build from there. In any case, my fix (adding rt to the link libraries in the qmake rules) for the build worked, so I guess we can just let the merge go.

Sorry if you weren't ready to have these changes merged yet!

andrew-elder:
Thanks, and merging is fine!
  
  
  

----------------------------------------

PR #235, opened 2015-06-06
Cross compile support

wgoossens:
Allow overriding the compiler/CC variable by using weak assignment to $(CC) and removing harcoded gcc entries. The maap, mrpd and mrp_control patches are cross-compile tested using a yocto build targetting a PPC and all changes are compile tested using native compile.


andrew-elder:
This looks good to me. Comments anyone else?

davidcemin:
Looks good to me too.

PR #238, opened 2015-06-10
Asi for upstream avb next

andrew-elder:
cmakefiles corrections and parse sscanf unsigned correction


andrew-elder:
Added more comments and check for NULL. I'm thinking I should create a unit test for this to be sure all is working as expected......

andrew-elder:
This is ready for merging if someone has time to review it. Changes are minimal and unit test has been added.

pinealservo:
I like the single bit tests, and they're probably sufficient here. In general, I like to test the values at the limits of the allowable range, though, as they're where edge cases tend to crop up.

I don't have any objections to the code as it is, but I'll hold off on merging in case you want to do something with the comments I've left.

andrew-elder:
Thanks for the detailed comments. I agree with them all and believe I have addressed them all as well.

PR #252, opened 2015-07-02
Fix bogus PHY latency values in gPTP Linux HAL

pinealservo:
On 2014-12-18 commit 5668e632c2063510b7f35e8a326c1622eab86325 introduced
a problem for users of the generic Linux HAL--some extremely large PHY
latency values (probably to offset latency introduced by a net sniffer)
were accidentally (I assume) committed. These cause low link delay values
to become negative.

This commit removes the preprocessor directives that replaced the original
latency values with the large ones.

### DIFF
--- a/daemons/gptp/linux/src/linux_hal_generic.cpp
+++ b/daemons/gptp/linux/src/linux_hal_generic.cpp
@@ -48,12 +48,8 @@
 #include <syscall.h>
 #include <limits.h>
 
-#if 0
 #define TX_PHY_TIME 184
 #define RX_PHY_TIME 382
-#endif
-#define TX_PHY_TIME 8000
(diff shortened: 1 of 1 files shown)


christopher-s-hall:
I would rather remove the PHY delay compensation code entirely and leave it up to the driver to do the "right thing".

davidcemin:
This patch solves the problem of the latency values.. I was doing that manually in our setup against extreme switch, but I agree with Chris that we should leave it up to the driver. 
I have no objections on merging this code to next/master but we should track this problem in another issue then.

pinealservo:
I'd like a better long-term solution too, but it is not a "one or the
other" thing. I can see putting them back where they were (this patch) or
perhaps removing them, but leaving them at 8000 makes no sense to me.

PR #259, opened 2015-07-24
Fixed memory leak where timer descriptor was not deleted; removed red…

christopher-s-hall:
…undant code

### DIFF
--- a/daemons/gptp/linux/src/linux_hal_common.cpp
+++ b/daemons/gptp/linux/src/linux_hal_common.cpp
@@ -213,6 +213,9 @@ void *LinuxTimerQueueHandler( void *arg ) {
 		    struct LinuxTimerQueueActionArg *arg = iter->second;
 			timerq->timerQueueMap.erase(iter);
 			timerq->LinuxTimerQueueAction( arg );
+			if( arg->rm ) {
+				delete arg->inner_arg;
+			}
 			timer_delete(arg->timer_handle);
 			delete arg;
 		}
@@ -267,7 +270,6 @@ bool LinuxTimerQueue::addEvent
 	outer_arg->rm = rm;
 	outer_arg->func = func;
 	outer_arg->type = type;
-	outer_arg->rm = rm;
 
 	// Find key that we can use
 	while( timerQueueMap.find( key ) != timerQueueMap.end() ) {


daemons/gptp/linux/src/linux_hal_common.cpp
```diff
@@ -213,6 +213,9 @@ void *LinuxTimerQueueHandler( void *arg ) {
 		    struct LinuxTimerQueueActionArg *arg = iter->second;
 			timerq->timerQueueMap.erase(iter);
 			timerq->LinuxTimerQueueAction( arg );
+			if( arg->rm ) {
+				delete arg->inner_arg;
+			}
 			timer_delete(arg->timer_handle);
 			delete arg;
```
andrew-elder: I'm confused. Doesn't delete arg; take care of deleting arg->inner_arg ?
Also, if the code is actually correct, I think it needs a comment as to why a test for arg->rm causes a delete of arg->inner_arg.

christopher-s-hall: No it doesn't (I don't understand why you've asked me this twice now).  There's a huge memory leak we've confirmed the problem and the fix.  Thanks.

andrew-elder: I'm going to accept this in 24 hours unless anyone else has comments. I'm sure the code is obvious to someone familiar to the code base. I was hoping you would suggest a comment along the lines of
// if rm (remove flag) is true, delete inner_arg
As a casual reader, I didn't even know rm was a flag rather than a pointer.

Sorry about asking the same question twice. I may have missed a response somewhere along the line.

andrew-elder: Thanks for this patch BTW.

pinealservo:
I'm okay with the fix as-is, since it's certainly a solution to the memory leak and doesn't appear to introduce any additional problems.  The full structure of `LinuxTimerQueueActionArg` is only available in this file, so the details of creating/deleting it are unlikely to cause trouble elsewhere.

It seems to me that it might be clearer and easier to maintain if the initialization and cleanup of LinuxTimerQueueActionArg were moved to a constructor/destructor pair, but I suppose that's arguable and probably best left to a separate commit if someone would like to propose it.

@andrew-elder, I think a comment regarding the reasoning behind the use of the 'rm' flag would indeed be helpful, but it would be most helpful as documentation to LinuxTimerQueue::addEvent(), which seems to be the only place where the flag is set, and it sets it directly from its parameter. But again, I'd suggest it as a separate documentation improvement commit.

davidcemin:
I concur. We should raise an issue for improving documentation on this case and then we can proceed with the merge the way it is.

PR #267, opened 2015-08-19
igb: fix build on 32-bit kernels

htfujina:
Fix build on 32-bit kernels.

### DIFF
--- a/kmod/igb/igb_ptp.c
+++ b/kmod/igb/igb_ptp.c
@@ -327,14 +327,12 @@ static int igb_ptp_adjtime_i210(struct ptp_clock_info *ptp, s64 delta)
 }
 
 static int igb_ptp_gettime64_82576(struct ptp_clock_info *ptp,
-				 struct timespec *ts64)
+				   struct timespec64 *ts64)
 {
 	struct igb_adapter *igb = container_of(ptp, struct igb_adapter,
 					       ptp_caps);
-	struct timespec ts;
 	unsigned long flags;
 	u64 ns;
-	u32 remainder;
 
 	spin_lock_irqsave(&igb->tmreg_lock, flags);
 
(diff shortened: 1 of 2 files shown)


andrew-elder:
There are some extra white space changes in the commit. Are they deliberate?

andrew-elder:
Can someone at Intel confirm/approve?

intel-ethernet:
Technical changes look good to me. I’ll let Todd comment on the whitespace.

htfujina:
The spacing fixes are to conform with the kernel standard (as I understand it). Not necessary, but the igb driver that this is based off of has those now as well.

andrew-elder:
Ok, if spaces come from somewhere upstream and are not "accidental" I'll accept the patch as is.

kmod/igb/igb_ptp.c
```diff
@@ -369,17 +365,14 @@ static int igb_ptp_gettime64_i210(struct ptp_clock_info *ptp,
 
 #ifdef HAVE_PTP_CLOCK_INFO_GETTIME64
 static int igb_ptp_settime64_82576(struct ptp_clock_info *ptp,
-				 const struct timespec *ts64)
+				   const struct timespec *ts64)
```
MarcinMiklasSTC: Why the parameter type wasn't changed here as in other functions? Shouldn't we have `const struct timespec64 *ts64` here?

htfujina: I think you're correct.

andrew-elder: @htfujina - will there be a new pull request to fix this then?

htfujina: Eventually, yes, but the history of timespec64 makes me think this won’t cause a failure and unless it does it’s not my highest priority. If nothing awful happens in the next week, I’ll have a pull request in a few days.

Todd Fujinaka
Software Application Engineer
Networking Division (ND)
Intel Corporation
todd.fujinaka@intel.com
(503) 712-4565

PR #268, opened 2015-08-19
Open avb next

MarcinMiklasSTC:
Couple of libigb fixes. Success returned on error from igb_attach_tx, missing locking in igb_clean.

### DIFF
--- a/lib/igb/igb.c
+++ b/lib/igb/igb.c
@@ -1411,3 +1411,129 @@ igb_set_class_bandwidth(device_t *dev,
 	return error;
 }
 
+int
+igb_set_class_bandwidth2(device_t *dev,
+	u_int32_t class_a_bytes_per_second,
+	u_int32_t class_b_bytes_per_second)
+{
+	u_int32_t	tqavctrl;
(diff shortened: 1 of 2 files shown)


andrew-elder:
Need review from Intel on this one.... I think it looks ok.

lib/igb/igb.c
```diff
@@ -1093,6 +1094,12 @@ igb_clean(device_t *dev, struct igb_packet **cleaned_packets)
 		if (txr->tx_avail >= IGB_QUEUE_THRESHOLD)	  
 			txr->queue_status &= ~IGB_QUEUE_DEPLETED;
 	}	
+
+unlock:
```
pinealservo: I don't see anything using this label in the function, and the conditional wrapping the sem_post() seems to not be serving any purpose either. I understand the desire to check return values, but it seems pointless to me to add a check that doesn't DO anything different based on the value. I guess it makes it obvious that it should be an early return if other code gets added on?

These concerns shouldn't affect the functionality of the code so I don't want to block it if it needs to go in and others think it's fine, but they did leave me scratching my head a bit as I reviewed the code.

PR #273, opened 2015-08-24
igb-avb: fix typo (timespec/timespec64)

htfujina:
Use timespec64 when necessary.


pinealservo:
If this is the change referenced in #267, it looks good to me.

PR #274, opened 2015-08-26
Feature avtp pipeline

MarcinMiklasSTC:
Rawsock implementation which uses libigb added. Frame size is currently limited to 256 bytes which is enough for AAF streams. For h264 streams it is usually not enough, you must change IGB_MTU constant. 

7 class A streams and 14 class B streams can be transmitted simultaneously without much problems. No crashes, only couple of late frames every 10 seconds.


andrew-elder:
@MarcinMiklasSTC - I merged your other pull request to open-avb-next. Does that mean this pull request should be modified?

MarcinMiklasSTC:
I will merge open-avb-next to this branch and cherry-pick other changes. Let's close it for now, because I see build error which needs to be checked.

MarcinMiklasSTC:
Patch was updated. It is now based on the merge from open-avb-next. The build error on  c5edda5
seems like a build script error. Am I right? So I guess It can be merged.

andrew-elder:
I need one other person to browse and sign-off on this pull request.

kencarlino:
I have reviewed the code changes and they look good to me. I do have one question related to the igb fixes that appear here in the feature-avtp-pipeline branch. Have these igb fixes also been applied to the open-avb-next branch? The source of my question is that I'm wondering if Intel has reviewed the igb changes since there is the expectation that the feature-avtp-pipeline branch will ultimately be merged into the open-avb-next branch.  Or perhaps the Intel review would happen at that time?.

andrew-elder:
@MarcinMiklasSTC, I believe you have had the policy of making igb changes on open-avb-next and then merging those changes in to feature-avtp-pipeline, correct? Either way, accepting this pull request does not prevent us from merging between open-avb-next and feature-avtp-pipeline at a later date.

MarcinMiklasSTC:
@andrew-elder Right, the changes to igb was done on open-avb-next and then merged to feature-avtp-pipeline.

Issue #277, opened 2015-08-28
Is there a way to change size of memory returned by igb_dma_malloc_page?


MarcinMiklas:
It only returns 4096 bytes of memory, which is not enough when I try to use it for use by multiple talkers.

It is enough for audio talkers, but h264 talker would like to use full ethernet frames, so in 4KB of memory I can only fit 2 of them, so that would limit the number of talkers to two.

I tried also to call igb_dma_malloc_page many times so that each talker has its own shared mem, but it turns out that it only works for 2 talkers. For 3 or more talkers I get bus errors, memory errors and whole machine crashes.

So I tried another approach to just call igb_dma_malloc_page twice so I get separate page for class A and class B streams.

And it happens to work quite well, allowing to run 7 class A streams or 14 class B streams without any crashes and moderate stream quality. With 4 class A streams quality is ok, that is that 8000 packets are sent per second (for 7 class A stream about 7950 packets are sent). All this tests were done with AAF streams, where packet size is 66 bytes.

So I would like to know if there is a way to increase number of bytes returned by igb_dma_malloc_page?

pinealservo:
Not without modifying both the igb library and the kernel module. The 'page' part of the name refers to the low-level pages of the virtual memory system, which have a fixed size. It's important to allocate at the page granularity because the memory for a buffer must be *physically* contiguous for DMA purposes rather than simply contiguous in a virtual address space, and the physical-to-virtual mappings happen at a page granularity.

It is possible in the kernel to allocate multiple contiguous pages, but it would require modifying the implementation of the IGB_MAPBUF ioctl in the kernel and the implementation of igb_dma_malloc_page(). The igb_mapbuf function in the driver would have to call alloc_pages() instead of alloc_page().

MarcinMiklasSTC:
Here is an attempt to fix that issue: #286

Now even 14 class A streams sending full frame works.

MarcinMiklasSTC:
Here is even better fix #295

PR #286, opened 2015-09-04
Changes tx buffer size from one page to 16 pages.

MarcinMiklasSTC:
Details: igb_mapbuf now allocates 16 pages instead of one,
using alloc_pages. This allows to increase the size
of tx packet to 1500 when multiple streams are sharing
the tx buffer.

### DIFF
--- a/kmod/igb/igb.h
+++ b/kmod/igb/igb.h
@@ -197,6 +197,10 @@ struct vf_data_storage {
 #define IGB_RX_BUFSZ	   IGB_RXBUFFER_2048
 #endif
 
+#define IGB_DMA_PAGE_ORDER 4
+#define IGB_DMA_PAGE_COUNT (1 << IGB_DMA_PAGE_ORDER)
+#define IGB_TX_BUFFER_SIZE ((PAGE_SIZE) * (IGB_DMA_PAGE_COUNT))
+
 
 /* Packet Buffer allocations */
 #define IGB_PBA_BYTES_SHIFT 0xA
--- a/kmod/igb/igb_main.c
+++ b/kmod/igb/igb_main.c
@@ -10189,27 +10189,27 @@ static long igb_mapbuf(struct file *file, void __user *arg, int ring)
 			adapter->userpages = userpage;
 		}
 
-		page = alloc_page(GFP_ATOMIC | __GFP_COLD);
+		page = alloc_pages(GFP_ATOMIC | __GFP_COLD, IGB_DMA_PAGE_ORDER);
 		if (unlikely(!page)) {
 			err = -ENOMEM;
 			goto failed;
 		}
-		
+
 		page_dma = dma_map_page( pci_dev_to_dev(adapter->pdev), page,
-						0, PAGE_SIZE,
+						0, IGB_TX_BUFFER_SIZE,
 						DMA_FROM_DEVICE);
-		
+
 		if (dma_mapping_error(pci_dev_to_dev(adapter->pdev), page_dma)) {
 	   		put_page(page);
 			err = -ENOMEM;
 			goto failed;;
 		}
-	   		 
+
 		adapter->userpages->page = page;
 		adapter->userpages->page_dma = page_dma;
 
 		req.physaddr = page_dma;
-		req.mmap_size = PAGE_SIZE;
+		req.mmap_size = IGB_TX_BUFFER_SIZE;
 	}
 
 	if (copy_to_user(arg, &req, sizeof(req))) {
@@ -10263,9 +10263,9 @@ static long igb_unmapbuf(struct file *file, void __user *arg, int ring)
 		if (userpage == NULL)
 			return -EINVAL;
 
-		dma_unmap_page(pci_dev_to_dev(adapter->pdev), 
+		dma_unmap_page(pci_dev_to_dev(adapter->pdev),
 				userpage->page_dma,
-				PAGE_SIZE,
+				IGB_TX_BUFFER_SIZE,
 				DMA_FROM_DEVICE);
 
 		put_page(userpage->page);
@@ -10343,9 +10343,9 @@ static int igb_close_file(struct inode *inode, struct file *file)
 	userpage = adapter->userpages;
 
 	while (userpage != NULL) {
-		dma_unmap_page(pci_dev_to_dev(adapter->pdev), 
+		dma_unmap_page(pci_dev_to_dev(adapter->pdev),
 						userpage->page_dma,
-						PAGE_SIZE,
+						IGB_TX_BUFFER_SIZE,
 						DMA_FROM_DEVICE);
 
 		put_page(userpage->page);


andrew-elder:
Are any experts of linux page allocation and linux DMA able to shed more light on this discussion?

MarcinMiklasSTC:
I can only add that we tested this fix for two days and we found no issues. 

Maybe it is not super optimal implementation but it works pretty well. We didn't notice any performance issues.

eric-mann:
I'm confused by what is being proposed or requested.

If you want (1) pinned 4K page, you call igb_dma_malloc_page() once. If you need more than (1) 4K page, you call igb_dma_malloc_page() as many times as you need 4K pages. If your app dumps core/abnormally exits, the driver will clean up and release any allocated pages. 

Is the question you want more than 4K - like 16K - of physically contiguous pinned memory? Or 16K of virtually (but not necessarily physically) contiguous memory? Or want to use large page mode (2MB pages)?

pinealservo:
@eric-mann He wants 16K of physically contiguous memory; he has modified igb_dma_malloc_page() to provide this by changing the ioctl implementation in the igb driver. It now uses alloc_pages() instead of alloc_page(), but still uses a single call to dma_map_page() except with a size of 16K instead of 4K. It seems to be working this way, but I am concerned it may be working by accident rather than due to correct use of dma_map_page().

eric-mann:
.. I suspect the change is technically "okay" but I'm still trying to understand the underlying usage model which is driving the change.

A 4K page can be divided into (2) 2K pages, more than enough to store standard IEEE 802.3 frames). 

What drives the requirement for a contiguous 16K page allocation vs (4) individual 4K page allocations? 

You aren't going to send jumbo frames with AVB (?). But even if you did - you can modify the igb usermode library to build a scatter-gather list to point to as many 4K physical buffers as you want (?).

You can also play games with the system page size by enabling huge pages and allocate far more than 16K (16K isn't a supported size by the way - https://www.kernel.org/doc/Documentation/vm/hugetlbpage.txt ).

Also - when we virtualize AVB and start using guest-physical-addressing (GPA) vs host physical addresses + IOMMU page registration, assumptions about "16K contiguous buffers" become an unwanted and system limiting (e.g. IOMMU translation cache size) complication unless there is a really good reason for it.

So I'm still confused.

pinealservo:
@eric-mann No doubt there are other ways to get this working, and probably much better ways, but this appears to be the easiest way to get the various assumptions made by the avtp-pipeline code about how things should work aligned with the assumptions made in the igb driver and library about how things should work.  Maybe scatter-gather is the 'right way' to do this, but the current library doesn't support it and I don't think we can expect @MarcinMiklasSTC to rework the driver and library to add that support. He's got a lot on his plate already!

The current igb library functions assume that you're going to be allocating buffers out of a contiguous chunk of memory, and his previous attempt to make multiple calls to igb_dma_malloc_page() caused problems (See previous discussion in #277). Maybe there's some other problem with how igb_dma_malloc_page() was being used, but I'm not familiar enough with that part of things to offer any concrete suggestions.

The avtp-pipeline code is a large chunk that's taken from a complete, working AVB code base at STC. It's got some assumptions about threading/processes/memory etc. that need to somehow get matched up with the various pieces we've got now in OpenAVB. I know you're busy, but this is really a great opportunity to move OpenAVB towards being a fully-functional AVB stack, and any effort that you could supply towards getting the interfaces matched up at the igb layer would go a long way at this point.

If you haven't got any time, I suggest we just merge what he's got working now and revisit later when you look at your AVB virtualization and GPA stuff. I know it's not a *great* solution, but the current igb library doesn't appear to provide a great solution, and this was a *simple* solution that only required small changes.

jdkoftinoff:
One option you have is to repeatedly call igb_dma_malloc_page() until you allocate enough contiguous pages.

--jeffk++

MarcinMiklasSTC:
@jdkoftinoff If I reserve more than two pages using igb_dma_malloc_page() and I try to use them then I get immediate crashes, because of bus errors in memcpy() ssl implementation and even whole PC freezes or reboots. This kind of bad stuff.

jdkoftinoff:
Why does it crash? Do you verify that the pages are actually contiguous? 

--jeffk++

eric-mann:
… is the issue the pipeline is expecting the physical pages are physically contiguous (e.g. paddr=0x1000, 0x2000, 0x3000)?
… or if the user-mode virtual addresses which point to the (possibly physically discontinuous) pages themselves are contiguous virtually (e.g. vaddr=0xa000, 0xb000, 0xc000)?
… are there other memory allocation requirements on the pages returned from the driver? (The Intel LAN cards are 64-bit capable – but some other devices are 32-bit (<4GB) or even low-memory (<16MB!) address limited.

Right now with the present design, allocation of 4KB pages one at a time means the neither the 4KB physical nor 4KB virtual addresses mapped into the process address space are guaranteed to be contiguous with other page allocations. If there is an application assumption that 1 page allocation maps more than 4KB of physical contiguous memory, the typical solution is to increase the page size – either 2MB (or even 1GB pages – used sometimes with virtualization).

ekm

jdkoftinoff:
I don’t know what the original issue is - But my suggestion is if for some reason contiguous physical memory addresses are required, it is possible to repeatedly call the allocate page function until it does return the required number of contiguous pages - possibly wasting a few pages initially, which can then be freed afterwards.  It is kind of ‘hackish’ but has been done before.

Regards,
Jeff

MarcinMiklasSTC:
I'll try to explain again what was the original problem.

I'd like to run *N* talkers. So I called `igb_dma_malloc_page()` *N* times, so that each talker has its own buffer to play with. And it worked ok for any values of *N* given that *N* is not bigger than 2 :|

So I do not need to build contiguous memory out of pages returned by `igb_dma_malloc_page()`. Separate buffer for each talker would be even better; there would be no need to play with mutexes to synchronize access to the buffer.

But for some unknown reason this approach doesn't work. With 3 talkers I get immediate crash, `bus error` mostly when trying to copy AVTP packet bytes to buffer returned by `igb_dma_malloc_page()`.

It seems that *libigb* or *igb_avb* module isn't prepared to work with more than two pages given to user.

That's why the idea to use one buffer from `igb_dma_malloc_page()` for each class came out and then this pull request which just gives more memory so that bigger frames can be sent.

eric-mann:
I’ll look at why you can’t allocate more than 2 pages … as it used to work (I explicitly tested the buffer allocation at one point long ago).

But if there is some reason to allocate bigger contiguous buffers for some reason … let me know as we can go in that direction as well. I just want to understand the usage behind it.

ekm

eric-mann:
Two requests: 

a) Can you post the "bus error" message you see (is it a kernel panic/oops? or a usermode signal SIGSEGV?)

b) are you allocating buffers from multiple threads? or also multiple buffers?

I'm suspicious of this code which I 'knew' was not thread safe when I wrote it ... and I only tested it in a version of simple_talker which just looped calling the api allocating buffers, then freeing them. I never tried multiple processes accessing the same interface+driver before ... 
 
		/* multi-threaded warning - a good impl woudl put a
		 * mutex around this
		 */
		if (adapter->userpages == NULL) {
			adapter->userpages = userpage;
		} else {
			userpage->next = adapter->userpages;
			adapter->userpages->prev = userpage;
			adapter->userpages = userpage;
		}

eric-mann:
s/multiple buffers/multiple processes/

MarcinMiklasSTC:
Please hold on with merging this pull request.

I don't experience crashes now. I need to carefully redo my tests.

andrew-elder:
Thanks for the update. Please keep us posted....

MarcinMiklasSTC:
I am very disappointed because I was not able to reproduce crashes and I do not understand why they disappeared :( Too many things changed in between, libigb, igb_avb, avtp_pipeline.

Good thing is that still with this pull request avtp_pipeline works better than without it (and separate buffer from `igb_dma_malloc_page()`  for each talker thread).

With 16 pages buffer shared between all talker threads I can run 4 class A streams with 1420 bytes of AVTP payload without any disturbances, I see that all talkers are able to send 8000 frames in each second.

In solution with separate 4K buffer for each talker thread I see that the same number of talkers (4) are not able to send even 8000 packets per second. I see numbers as low as 7990 most of the time. Even worse at the beginning of transfer I see more than 200 `igb_xmit failed: No space left on device` messages. Then talkers seems to auto synchronize and the igb_xmit failurs happen less often, but they are still there.

For my testing I turned off launch time feature in driver (pull request in preparation), and I am not calling `igb_get_wallclock()`. All this to just test pure packet sending.

I've noticed interesting issue with this pull request and simple_talker example. It doesn't work correctly sending bursts of noise instead of continuous sine wave.  To fix it you must set a_page.mmap_size to value less than 12900 or you packet_size so that a_page.mmap_size/packet_size <= 128. I have no clue why this is happening. 128 is the magic value which suggest that there can be some 7 bit variable keeping track of used buffers which get overflow when bigger value is written there or something like that (but then 127 should be a limit).

I'll put the other implementation on the branch in my repo.

eric-mann:
I can only imagine the "no space left on device" error (where ever it is coming from) means you have run out of transmit descriptors with posted transmit packets (or you don't clean up often enough). You can 'fix' this by increasing the available transmit descriptors when you start the igb_avb driver. 

I can't comment about your performance issues. The device can send 64-byte packets at 1Gbps. 

There are no '7 bit' variables in the igb driver or associated library ...  this sounds like something related to your 16KB buffer allocation modification (?). 

I don't agree with merging the changes to the memory allocation. I suggest you focus on improving the performance of your memory management routines - perhaps use a lockless design partitioning buffers to CPU cores or somesuch like commonly done in the kernel.

MarcinMiklasSTC:
I've created two branches with two approaches for multiple talker:

https://github.com/MarcinMiklasSTC/Open-AVB/tree/avtp-pipeline_64kbuf

https://github.com/MarcinMiklasSTC/Open-AVB/tree/avtp-pipeline_nosharedbuf

You can find there ./test.sh script which can be used for testing. It takes up to 4 parameters:
 - ifname
 - number of streams (1-16)
 - payload size (20-1400)
 - class (A,B)

Using this script shows how much the *avtp-pipeline_64kbuf* approach is better. You can run without glitches:
 - 16 class A streams with 600 bytes of payload
 - 7 class A streams with 1400 bytes of payload
 - 16 class B streams with 600 bytes of payload
 - 15 class B streams with 1400 bytes of payload

For comparison *avtp-pipeline_nosharedbuf*. You can run without glitches:
 - 2 class A streams with 600 bytes of payload
 - 1 class A stream with 1400 bytes of payload
 - 2 class B streams with 600 bytes of payload
 - 1 class B stream with 1400 bytes of payload

I hope that these measurements will convince you that this pull request is worth merging. I know that the things can be done better, but it is rather simple change which gives amazing results.

There is still possible that I made some stupid mistake in my code, that's why I published my working branches. Please redo my experiments yourself and try your own improvements.

andrew-elder:
@eric-mann, is anyone on your side available to do a more detailed code review? We seem to be approaching an impasse here....

MarcinMiklasSTC:
Will leaving behaviour of igb_dma_malloc_page() intact and adding separate igb_dma_malloc_page_ex() which takes order of pages as a parameter be more acceptable approach?

I see here the classic theory vs practise dilemma. In theory we know why it shouldn't work but it practise it works great but we have no idea why :)

MarcinMiklasSTC:
Lets close this pull request.

I am now testing new implementation of avtp_pipeline rawsockets which creates big chaing of tx_frames from memory returned by multiple calls to igb_dma_malloc_page(). 

Only difference is that some memory is wasted. I need 1522 bytes for full ethernet frame with vlan tag.  page_size % 1522 = 1052, so I can fit only 32 frames in 16 pages of allocated memory  (in continuous 64KB buffer I can fit 42 frames).

Thanks for all the comments.

PR #288, opened 2015-09-09
Added function igb_get_mac_addr().

MarcinMiklasSTC:
libigb reads mac address from hardware but doesn't provide function to access it. This pull request adds such function.


eric-mann:
Looks fine to me.

PR #289, opened 2015-09-10
SRP integration

MarcinMiklasSTC:
Added communication with *mrpd* daemon so that dynamic streams reservation can be done.

To enable this pass -DAVB_FEATURE_ENDPOINT=1 parameter to cmake.


lib/avtp_pipeline/openavb_common/mrp_client.c
```diff
@@ -203,11 +209,37 @@ int process_mrp_msg(char *buf, int buflen)
...
+
+		sscanf(&(buf[i]), "V=%d,Z=%d,I=%d,P=%d,L=%d",
+		       &vid,
+		       &max_frame_size,
+		       &max_interval_frames,
+		       &priority_and_rank,
+		       &latency);
+
+		mrp_register_cb(recovered_streamid, 0, dest_addr, max_frame_size, max_interval_frames, vid, latency);
+
```
andrew-elder: If I were messing with messages from mrpd, I would use this code https://github.com/AVnu/Open-AVB/blob/master/examples/mrp_client/mrpdhelper.h

MarcinMiklasSTC: When I found mrpdhelper.h I had most of integration done, so I chose to continue with my approach. 

Current code is not pretty but it works. I can refactor code to use mprdhelper.h in separate pull request.

andrew-elder:

davidcemin:
Looks good to me too.

PR #290, opened 2015-09-11
Added igb_control_launch_time()

MarcinMiklasSTC:
This can be used to disable launch time feature of IGB
(driver enables it by default).

Change-Id: I94d3550feac505fea1c2c14e20c817aa575065ea

### DIFF
--- a/lib/igb/igb.c
+++ b/lib/igb/igb.c
@@ -1552,3 +1552,33 @@ int igb_get_mac_addr(device_t *dev, u_int8_t mac_addr[ETH_ADDR_LEN])
 	return 0;
 }
 
+int igb_control_launch_time(device_t *dev, int enable)
+{
+	struct adapter	*adapter;
+	struct e1000_hw *hw;
+
+	if (NULL == dev) return EINVAL;
+	adapter = (struct adapter *)dev->private_data;
+	if (NULL == adapter) return ENXIO;
+
+	hw = &adapter->hw;
+
+	if( sem_wait( adapter->memlock ) != 0 ) {
+		return errno;
+	}
+
+	unsigned int regVal = E1000_READ_REG(hw, E1000_TQAVCTRL);
+
+	if (enable)
+		regVal |= E1000_TQAVCTRL_LAUNCH_VALID;
+	else
+		regVal &= ~E1000_TQAVCTRL_LAUNCH_VALID;
+
+	E1000_WRITE_REG(hw, E1000_TQAVCTRL, regVal);
+
+	if( sem_post( adapter->memlock ) != 0 ) {
+		return errno;
+	}
+
+	return 0;
+}
--- a/lib/igb/igb.h
+++ b/lib/igb/igb.h
@@ -102,6 +102,8 @@ int igb_unlock( device_t *dev );
 
 int igb_get_mac_addr(device_t *dev, u_int8_t mac_addr[6]);
 
+int igb_control_launch_time(device_t *dev, int enable);
+
 #endif /* _IGB_H_DEFINED_ */
 
 


lib/igb/e1000_defines.h
```diff
@@ -1796,6 +1796,7 @@
 #define E1000_TQAVCTRL_FETCH_TM_MASK   0xFFFF0000 /* fetch time delta offset from launch */
 #define E1000_TQAVCTRL_FETCH_TM_SHIFT  (16)       /* ... and associated shift value */
 
+#define E1000_TQAVCTRL_DATA_TRAN_TIM   0x00000200 /* data launch time valid */
```
andrew-elder: Repeat of #define at line 1794 of same file.

lib/igb/e1000_regs.h
```diff
@@ -152,6 +152,9 @@
 #define E1000_INVM_DATA_REG(_n)	(0x12120 + 4*(_n))
 #define E1000_INVM_SIZE		64 /* Number of INVM Data Registers */
 
+/* QAV Tx mode control register */
+#define E1000_I210_TQAVCTRL	0x3570
+
```
andrew-elder: Repeat of #define at line 542 of same file.

eric-mann:
I have a technical counter argument with this (modify the igb_xmit() routine to not include the launch time context descriptor ... saves descriptors)... but I think the larger question is how do you intend to use the Qav shaper?

If you enable *only* the Qav shaper and not use any of the launch times for any stream, there will be a race condition on the talker side dynamically adjusting the Qav shaper bandwidth while pending stream transmissions are active. For example, assume you have Class A streams 1 & 2 reserved and active, then you get a listener for Class A stream 3, you (a) need to increase the Qav bandwidth and then (b) start sending stream 3 packets. In the window between (a) and (b) packets for streams 1 & 2 will go out at a higher than expected rate (if you have posted several packets ahead of the current time).

Is this what you intend?

kencarlino:
@eric-mann, The AVTP pipeline will not post packets ahead of time. The AVTP pipeline places packets into the igb xmit buffer at the interval based on the SR Class only when they are expected to be sent as soon possible. That is why the i210 launch time can be disable in this case. In the Window between (a) and (b) if there are packets pending in the tx queue awaiting credit it would seem it is okay that they will be sent based on the new CBS allocation. Once the 3rd stream packets start getting submitted for transmission the shaper will adjust.

andrew-elder:
This added API looks ok to me. Arguments as to whether or not it is a good idea to use it are application level decisions in my head at least. Any further comments? Am I missing something else? Is there a strong argument that applications should ALWAYS use launch time context descriptor? I think in this case the goal is actually to turn off the launch time feature.

kencarlino:
The compelling reason to have the option to not use packet launch times is for future portability. Not all MAC hw/sw support launch times. In my experience most MAC hw/sw don't have this kind of launch time feature.

eric-mann:
I know license-able soft IP is available from vendors which include the 'launch time' capability. I don't have any direct knowledge of other Ethernet vendors implementations, but from what I understand about the requirements of OS'es which natively support AVB, any AVB-capable Ethernet controller must have launch time support (?).

I'd rather modify the igb_xmit() routine to pass an additional parameter 'launch_valid' which the routine can decide to skip over generating the additional and unnecessary context descriptor than just disabling 'launch_valid' (which we only use as a test feature in practice).

jdkoftinoff:
I agree, Eric.

Definitely your recommendation would work for architectures where the launch time of the frame is directly controlled by a different method, such as a separate dedicated CPU core.

Jeff

MarcinMiklasSTC:
libigb is library for communicating with igb_avb.ko which is a driver for i210 which has register for controlling if launch time is valid. This simple pull request just adds the function which allows to change the register. 

This function really helps in feature-avtp-pipeline branch as AVTP pipeline isn't prepared for using launch time feature (yet). 

If for some reason you don't like this function added (like there is enough functions in API already), I can implement the functionality in application using igb_readreg/igb_writereg.

andrew-elder:
@MarcinMiklasSTC was there are specific reason you didn't comment on @eric-mann's suggestion to modify igb_xmit() to support not including a launch time descriptor? I think Eric is saying that he recommends the i210 ALWAYS be run in a mode that supports launch time. It a specific packet does not wish to use the launch time feature, then don't include a launch time descriptor. I'm glossing over some API details here, but would still appreciate your feedback.

MarcinMiklasSTC:
@andrew-elder Well I didn't comment on that because I have no time thinking about how to implement this. Just adding function to just change bit in I210 register is much simpler than that. I still have little experience in libigb and igb_avb. 

Maybe one of more experienced guy will try to implement that. I can help with testing.

MarcinMiklasSTC:
Suppose we have added `attime_valid` parameter to `igb_xmit()`. What if we send one frame with valid attime and then another one (from different talker) with invalid attime? Will the second frame be sent immediately or will the first packet be sent first and just after it the second one? The first option is expected one, but will it happen?

When I was playing with various AVTP pipeline implementations using launch time I noticed that the list of packets to be sent is not sorted by attime as it ought to be (should application do that???).  So if one talker sends frame with attime 1000 and then second talker sends frame with attime 900 then frame from second talker will be sent after time 1000 instead of 900.

Because of that I was not able to run more than one talker using launch time feature.

Also I didn't see any improvement in frames sending (I observed timestamp deltas between packets received on listener) using launch time feature and not using it. In both cases timestamp deltas was between 124 and 126 us (1 us is the timestamp resolution in wireshark on listener).

eric-mann:
Inline …

ekm


Suppose we have added attime_valid parameter to igb_xmit(). What if we send one frame with valid attime and then another one (from different talker) with invalid attime?

Point of clarification - there is no such thing as an ‘invalid’ attime. Either the supplied attime refers to something in the ‘recent past’ (< 0.5 ago), or attime refers to sometime in the ‘near future’.



Will the second frame be sent immediately or will the first packet be sent first and just after it the second one? The first option is expected one, but will it happen?

The I210 HW queues are first-in-first-out. So if SW supplies packets ‘out of order’ chronologically, the packets will be transmitted in-order as posted to the HW. However, the packets posted to HW later (but to be sent earlier) would appear to come out immediately (subject to Qav traffic shaping requirements).

When I was playing with various AVTP pipeline implementations using launch time I noticed that the list of packets to be sent is not sorted by attime as it ought to be (should application do that???). So if one talker sends frame with attime 1000 and then second talker sends frame with attime 900 then frame from second talker will be sent after time 1000 instead of 900.

SW has to sort packets in-order before posting to HW (as you observed).

Also I didn't see any improvement in frames sending (I observed timestamp deltas between packets received on listener) using launch time feature and not using it. In both cases timestamp deltas was between 124 and 126 us (1 us is the timestamp resolution in wireshark on listener).

Not sure what the metric for characterizing “improvement” would be. I could write a book on the sources of Ethernet transmission jitter.  For your usage – where multiple streams are comingled onto the same HW queue and a module doesn’t exist to sort and schedule packets onto HW, it may be better to not use the launch time and rely solely on the Qav shaper.

MarcinMiklasSTC:
@eric-mann Thanks for the detailed explanation. And thanks for confirming also my conclusion that launch time feature is not very handy for multiple talkers use case.

andrew-elder:
I think whether or not launch time is useful depends on the threading design decisions. If there is an independent and un-coordinated thread per talker, using launch time might not be useful. But if a single thread is responsible for packetization and queuing packets for transmission, one could image how having a programmed launch time might allow a reduction in the processor interrupt rate by submitting packets for multiple time slots - of course delay is increased in this case.

andrew-elder:
@MarcinMiklasSTC, do you have have further pending updates to this pull request?

MarcinMiklasSTC:
@andrew-elder No, I don't have. I understand why changing igb_xmit() function could be better but I don't have enough experience in libigb to do it. 

In feature_avtp_pipeline I implemented the same thing without extending libigb (just using function igb_reg_read/igb_reg_write).

andrew-elder:
@MarcinMiklasSTC, shall we close this pull request then? We could instead create a feature request to extend igblib to optionally support packet launch times.

PR #291, opened 2015-09-15
gptplocaltime() function added.

MarcinMiklasSTC:
It can be used to quickly get localtime calculated from current system time and values provided by gptp deamon (frequency offsets and phase offsets).

### DIFF
--- a/daemons/gptp/linux/src/ipcdef.hpp
+++ b/daemons/gptp/linux/src/ipcdef.hpp
@@ -47,7 +47,7 @@ typedef struct {
 	int64_t ls_phoffset;			//!< Local to system phase offset
 	FrequencyRatio ml_freqoffset;	//!< Master to local frequency offset
 	FrequencyRatio ls_freqoffset;	//!< Local to system frequency offset
-	int64_t local_time;				//!< Local time of last update
+	uint64_t local_time;				//!< Local time of last update
 	uint32_t sync_count;			//!< Sync messages count
 	uint32_t pdelay_count;			//!< pdelay messages count
 	PortState port_state;			//!< gPTP port state. It can assume values defined at ::PortState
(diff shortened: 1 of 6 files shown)


christopher-s-hall:
Good idea overall.  One niggle.

Why is it necessary to add system time to "update" structure?  System time can always be derived from local time using the phase offset.  I would be fine with using system time *instead* of local time.  I think we should avoid cluttering the structure more than it already is.

jdkoftinoff:
Wait - System Time is acquired via  "if (clock_gettime(CLOCK_REALTIME, &sys_time) != 0)”

CLOCK_REALTIME is not monotonic. Perhaps a better option is to use CLOCK_MONOTONIC_RAW, which is not even affected by any NTP rate adjustments.

We have to make sure that discontinuities in gPTP Grand Master time (because of changed grandmaster etc) and that discontinuities between current time of day (which can be affected by leap seconds and daylight savings time etc) are properly handled.

Regards,
Jeff

christopher-s-hall:
Completely agree with you, Jeff, on appropriateness of monotonic raw, however, the current Linux mechanism for getting the cross time-stamp is:

ioctl( phc_fd, PTP_SYS_OFFSET, &offset );

The "system" time is the REALTIME clock.  This is Linux kernel design decision over which we have little control.

jdkoftinoff:
Thank you for the clarification

That really is not good. 

I understand that we have to do it this way, but what are the possibilities of getting a new ioctl flag added to the Linux kernel, say PTP_MONOTONIC_RAW_OFFSET ??

Regards,
Jeff

MarcinMiklasSTC:
@christopher-s-hall You have right, I don't have to extend shared gPtpTimeData structure as the system_time can always be calculated from local time and ls_phoffset.

MarcinMiklasSTC:
I've updated commits. Now the gPtpData  doesn't have system_time field added.

MarcinMiklasSTC:
Hold on with merging, I must do additional tests.

MarcinMiklasSTC:
Now it is all ok.

christopher-s-hall:
Agree on both counts.  Looks good to me.

PR #304, opened 2015-10-09
documents: design: add avb_ipc.asciidoc draft

andrew-elder:
start of IPC doc

### DIFF
--- a/documents/design/oavb_ipc.asciidoc
+++ b/documents/design/oavb_ipc.asciidoc
@@ -1,0 +1,83 @@
+= Interprocess Communication Recommendations for AVB/TSN
+Andrew Elder
+v0.1
+
+== Introduction
+
+The AVB/TSN software stack is comprised of a number of modules that implement protocol stacks specified by IEEE standards. Each of these protocols implement an Ethernet network facing interface and an application facing interface. The required application facing interface is described in some detail in http://avnu.org/wp-content/uploads/2014/05/AVnu_SWAPIs_v1.0.pdf[AVB Software Interfaces and Endpoint Architecture Guidelines]. Since a module running a particular protocol may be a standalone execution unit (for example a Linux daemon), an interprocess communication method is required to interface to the execution unit.
+
+For portability reasons it is recommended to separate the implementation of the interprocess communication method as much as possible from the logic of the protocol implementation. This applies to both an application that is interfacing to an AVB/TSN module and the implementation of the AVB/TSN module itself.
+
+For the purposes of this document the following straw-man application interfaces are described:
+
+.Straw-man Interfaces
+[source,c/c++]
+----
+
+void AVBTSN_SetParam(uint32_t param);
+
+----
+
+where the AVBTSN portion of the function name would be replaced by the name of the IEEE AVB/TSN protocol module that is being targeted.
+
+
+== Communication Stack Abstraction
+
+This section outlines abstraction requirements for multiple layers. The layers that will be described here are represented in the following table.
+
+[width="85%",options="header"]
+|=======
+|Layer |Description
+|Application |Call APIs for AVB/TSN module
+|Packing |Takes parameters from API calls and packs them for "transport"
+|Communication |Send packed parameters
+|Channel |Transfer information from source to destination
+|Communication |Receive packed parameters
+|Unpacking |Unpacks information from channel
+|Module |This is the layer with equivalent calls to the top Application layer
+|=======
+
+=== Application Layer
+
+The application layer requires a simple interface that has parameters that closely match those supported by the underlying module implementation. However, since the context for an underlying communication layer is required, an additional IPC context parameter must be added to the application's calling interface. The example straw-man interface now becomes:
+
+.Straw-man Interfaces for External Application
+[source,c/c++]
+----
+
+void AVBTSN_SetParam(void *ipc_context, uint32_t param);
+
+----
+
+=== Packing Layer
+
+The marshalling layer is responsible for encoding parameters into a defined structure for passing through the communication layer.
+
+=== Abstract Communication Layer
+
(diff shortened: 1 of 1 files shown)


documents/design/oavb_ipc.asciidoc
```diff
@@ -0,0 +1,83 @@
+= Interprocess Communication Recommendations for AVB/TSN
+Andrew Elder
+v0.1
+
+== Introduction
+
+The AVB/TSN software stack is comprised of a number of modules that implement protocol stacks specified by IEEE standards. Each of these protocols implement an Ethernet network facing interface and an application facing interface. The required application facing interface is described in some detail in http://avnu.org/wp-content/uploads/2014/05/AVnu_SWAPIs_v1.0.pdf[AVB Software Interfaces and Endpoint Architecture Guidelines]. Since a module running a particular protocol may be a standalone execution unit (for example a Linux daemon), an interprocess communication method is required to interface to the execution unit.
```
pinealservo: I'd strike 'Ethernet'; although Ethernet is the only MAC layer I'm aware of with solid support for AVB/TSN so far, it's definitely not the only one in the standards.

documents/design/oavb_ipc.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+
+.Straw-man Interfaces
+[source,c/c++]
+----
+
+void AVBTSN_SetParam(uint32_t param);
+
+----
+
+where the AVBTSN portion of the function name would be replaced by the name of the IEEE AVB/TSN protocol module that is being targeted.
```
pinealservo: One difficulty that can arise with get/set-based interfaces is the situation when some set of changes need to be made more-or-less simultaneously across a set of parameters. You typically need some way to encapsulate a complete transaction or provide the same sort of functionality at a lower level, such as a 'shadow' block of parameters along with a 'trigger' that commits the shadow block to the actual configuration.

Do you intend for this interface to use a uniform datatype like uint32_t for parameters, or will the parameter type vary by parameter? What about communicating failure to get/set a parameter? Will responses to commands be communicated synchronously or asynchronously, or will it be up to the implementation?

Some of the protocol modules can expose quite a few parameters, and in some cases duplicates of the same set of parameters over multiple instances of some interface. E.g. ports or streams. It might make sense to impose a hierarchy on the interface identifiers, at least at the abstract interface description level.

andrew-elder: Thanks for the feedback. In my opinion a "set" of parameters would be communicated by adding multiple parameters to the parameter list and then bundling them internally in a structure. I'll add an example showing that since it is obviously not clear at this stage. I intend datatypes to vary on a parameter by parameter basis.

Hmmm - have to this about error handling. Ideas?

Can you elaborate on the hierarchy of interface identifiers idea? I was thinking each functional operation would map internally to a unique ID in the packing/unpacking layer that would support correct handling of the transaction, but I'm not sure if that is what you are referring you. I need to add more text. Trouble is it is hard to document how it might work without generating a concrete example.

documents/design/oavb_ipc.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+
+
+== Communication Stack Abstraction
```
pinealservo: I think this is a reasonable set of layers, but it might be a good idea to decide whether there's a desire to have implementations be compatible with one another and at which layer they ought to be compatible. I imagine the choice could be made differently from platform-to-platform based on the software vendors in that area and the desire for cross-vendor compatibility.  Also, some layers may or may not be necessary depending on the level of integration between components.

In some situations (and I think it's definitely a "best practice" to implement things this way) a protocol module will define all its upper-layer interfaces first as a natively-callable library interface. Then they would provide a wrapper application that provides the IPC layer for when the module is desired to run as a separate process. And on a similar note, not all IPC layers will require data packing/unpacking, and there's a significant overhead to performing that kind of data transformation. I think the layering interfaces ought to provide a more-or-less 'free' way of skipping layers that aren't required on a given platform/IPC combination. I'm guessing you've thought of this already, but I think it's good to make it explicit. It would be nice if the wrapper applications could be written in a standardized way, relying on the same set of libraries to provide the lower-level layering that is required and allowing the layering to be easily swapped for different platforms.

My main concern here is that we provide guidance that's reasonable for the range of platforms from Autosar-based automotive platforms (which take a very synchronous/time-triggered approach to things and require interfaces to have relatively low abstraction overhead) to Linux and Windows platforms (these can tolerate a reasonable amount of abstraction and interface flexibility) and then back again to standalone speakers or sensors/motor controllers with perhaps no OS at all.

documents/design/oavb_ipc.asciidoc
pinealservo: If you're going to do multiple parameters and a synchronous interface, a reasonable error mechanism is to always pass output parameters via 'out' argument pointers and use the function return value exclusively as a status code. For more detailed errors, provide an additional API function to retrieve module-specific error data via out parameters.

For an asynchronous interface, you would provide both a success callback function and an error callback function when making an API call. The nice thing about callbacks is that they can pass their parameters by value instead of requiring out-parameter pointers. The not-so-nice thing is how you have to chop your functions into several pieces and put common variables in a structure that's not stack-allocated.

I'd try to avoid specifying anything about allocation so that embedded devices can use static rather than heap allocation for context/state structures.

documents/design/oavb_ipc.asciidoc
davidcemin: I agree with Levi, but I think it might be hard to generalize one implementation guideline throughout all the different platforms we are using open-avb on. What about having one section for general guidelines, with topics that would apply for all platforms and platform-specific sections detailing a little more what would be a best practice for that specific platform?

andrew-elder: I was thinking about having a concrete examples section that would highlight synchronous and asynchronous communication options. Thinking about it some more, it doesn't make much sense to talk about layers in a general section that are not required, so perhaps I will restructure the whole document around synchronous and asynchronous calling options. I sometimes call those options blocking and non-blocking. I'll make that change.

PR #305, opened 2015-10-14
Feature ipc doc

andrew-elder:
split in to sync and async sections


davidcemin:
Since its a WIP im merging it in so other people can work in parallel. Any suggestions can still be written here and we update it on the next pull request.

PR #324, opened 2015-11-20
Added support for Intel wireless adapters on Windows

christopher-s-hall:
Moved port states to ipcdef
Added additional simpler debug output that always prints used for compile
	time #ifdef wrapping
At startup when in slave only or "cached slave" mode set restart
	syntonization flag
Added wireless sync specific code for non-GM multiport operation
Delete pdelay timeout in the default handler preventing asCapable from
	becoming false on links that don't use pdelay
Remove rollover handling from common wireless timestamper code

Removed broken functions/macros for timestamp arithmetic replacing them with
	"C++ style" operator overloading
Use two wireless interface addresses, one for media independent functions
	the other for media dependent functions
	No longer necessary to "hard-code" transform from wireless hardware
		address to ethernet wireless address (may be different for
		WiFi Direct)
Removed unnecessary TSC complication, instead use
	QPC frequency << 10 <=> TSC frequency
Added work queue for "true" asynchronous behavior
Split wired and wireless windows code
Include updated solution/project files
Pulled Linux code from the main branch enabling use of PTP driver
	cross-timestamp


andrew-elder:
Mostly looks good to me. A couple of small correction requests
1) some of the files don't end a blank line (github has a red error at the end of the file in the "Files changed" list.
2) windows_hal_wireless.cpp has INTEL_XXXX defines in it. Should it be called intel_hal_wrieless.cpp perhaps?

andrew-elder:
Who else is able to comment on this code?

daemons/gptp/common/ptp_message.cpp
```diff
@@ -803,9 +805,13 @@ bool PTPMessageFollowUp::processMessage(MediaDependentPort *port, Timestamp rece
 		-(TIMESTAMP_TO_NS(corrected_sync_time - system_time));
 
 	offset.master_time = TIMESTAMP_TO_NS(preciseOriginTimestamp);
+	//printf("Master Time: %llu\n", offset.master_time);
+	//printf("System Time: %llu\n", TIMESTAMP_TO_NS(system_time));
```
davidcemin: Do we need to keep this commented code? Do we already have a standardized of way of enabling/disabling debugs?

daemons/gptp/common/wltimestamper.hpp
```diff
@@ -83,18 +83,8 @@ class WirelessDialog {
...
-		*rollover = *rollover + 1;
-	}
-	ret = *rollover;
-	ret <<= 32;
-	ret += curr;
-	*prev = curr;
-	return ret;
-}
-
+#define MAX_NSEC (42949672960)
```
davidcemin: Isnt this already defined on limits.h on the macro ULONG_MAX ?

davidcemin:
My comments are minor and the code looks good to me.

andrew-elder:
@christopher-s-hall - are you able to respond to the above comments sometime this week?

daemons/gptp/common/wltimestamper.hpp
pinealservo: ULONG_MAX would be 4294967295 or 0xFFFF_FFFFU, so MAX_NSEC (42949672960, or 0x0000_000A_0000_0000ULL) is 10*(ULONG_MAX + 1).  The parameter being passed to it is being multiplied by 10, so it's possibly a bugfix, but with the weird MAX_NSEC and the multiplication by 10 and the _32to64(uint64_t) function, the section could stand some refactoring or at least explanatory comments.

christopher-s-hall:
I incorporated David's comments verbatim by either making the suggested change or removing the offending code entirely.  For Andrew's comments, I fixed the newlines and reworked the code some.  I moved the Intel HW specific code into intel_wireless.*.  The idea is that windows_hal_wireless.* should be fairly generic.  I believe that is now the case.  Moved structures that should be OS common into the common directory.

daemons/gptp/common/md_wireless.cpp
```diff
@@ -231,7 +240,7 @@ bool MediaDependentWirelessPort::processSync(uint16_t seq, bool grandmaster, lon
 				unsigned residence_time;
 				follow_up->setPreciseOriginTimestamp
 					(getClock()->getLastSyncOriginTime());
-				// Calculate residence time
+				// Calculate residence time  XXX : This looks suspicious as if residence time could be negative
```
pinealservo: Could you elaborate on your concern in this comment?

daemons/gptp/common/wltimestamper.hpp
```diff
@@ -157,7 +212,21 @@ class WirelessTimestamper : public Timestamper {
...
+		if (curr < last_count && (last_count - curr) >(MAX_NSEC >> 1)) {
+			rollover = rollover + 1;
+		}
+		ret = rollover;
+		ret *= MAX_NSEC;
+		ret += curr;
+		last_count = curr;
+		rollover_lock->unlock();
+		return ret;
+	}
```
pinealservo: The redefinition of MAX_NSEC definitely made it clearer why it was defined to the value it was. I think this function, `uint64_t _32to64(uint64_t curr)`, could use a better name. Maybe something to do with rollover handling.

As a general question, and not something you necessarily need to resolve:  Is it correct that the source of the value you're working with here is expressed in tens of nanoseconds (i.e. each tick is 10ns) and comes from a 32-bit register? Is this scaling factor part of the WiFi spec, or is it something that could differ between WiFi chipsets?  It might improve clarity to separate out the rollover management from the scaling-to-ns process, but I haven't thought about it deeply enough to be sure.

christopher-s-hall: The idea is that the driver doesn’t do the “right thing” and this is a kludge until it is fixed.  I guess I would ask that since this is not a main-line branch and that it is being released to “spur” development within AVnu that you disregard this for now.

daemons/gptp/common/md_wireless.cpp
christopher-s-hall: I’ll remove the comment, it may have made sense at one time, but not now.

andrew-elder:
I think Chris has addressed all concerns at this point. Do others agree?

davidcemin:
I do. Thank you Chris.

PR #328, opened 2015-12-08
New: VLAN ID can be configured for no endpoint builds.

MarcinMiklas:
Details: You can specify VLAN Identifier in talker ini file.
Syntax is vlan_id = 5; Default value is 2 for both class A and class B streams.

### DIFF
--- a/lib/avtp_pipeline/tl/openavb_talker_no_endpoint.c
+++ b/lib/avtp_pipeline/tl/openavb_talker_no_endpoint.c
@@ -74,12 +74,14 @@ bool openavbTLRunTalkerInit(tl_state_t *pTLState)
 	pTalkerData->streamID.uniqueID = pCfg->stream_uid;
 	if (pCfg->sr_class == SR_CLASS_A) {
 		pTalkerData->classRate = 8000;
-		pTalkerData->vlanID = SR_CLASS_A_DEFAULT_VID;
+		pTalkerData->vlanID = pCfg->vlan_id == VLAN_NULL ?
+					SR_CLASS_A_DEFAULT_VID : pCfg->vlan_id;
 		pTalkerData->vlanPCP = SR_CLASS_A_DEFAULT_PRIORITY;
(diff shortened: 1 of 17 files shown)


lib/avtp_pipeline/tl/openavb_talker_no_endpoint.c
```diff
@@ -74,12 +74,14 @@ bool openavbTLRunTalkerInit(tl_state_t *pTLState)
...
 		pTalkerData->classRate = 8000;
-		pTalkerData->vlanID = SR_CLASS_A_DEFAULT_VID;
+		pTalkerData->vlanID = pCfg->vlan_id == UINT16_MAX ?
+					SR_CLASS_A_DEFAULT_VID : pCfg->vlan_id;
 		pTalkerData->vlanPCP = SR_CLASS_A_DEFAULT_PRIORITY;
 	}
 	else if (pCfg->sr_class == SR_CLASS_B) {
 		pTalkerData->classRate = 4000;
-		pTalkerData->vlanID = SR_CLASS_B_DEFAULT_VID;
+		pTalkerData->vlanID = pCfg->vlan_id == UINT16_MAX ?
```
andrew-elder: This is a little confusing? I assume the intent is that UINT16_MAX is used as VLAN_NULL?
It would be clearer to have
#define VLAN_NULL UINT16_MAX
in a header file and then use VLAN_NULL in the source code.
Or, you could add a comment here to say that UINT16_MAX is used to represent a NULL value.

andrew-elder:
I have one small comment. Otherwise it looks good to me. Any other comments?

MarcinMiklas:
Added and used VLAN_NULL define.

andrew-elder:
Thanks @MarcinMiklas for adding this. All looks good to me. I need one other person to sign-off and/or review and accept the pull request.

davidcemin:
Looks good to me.

PR #330, opened 2015-12-14
Refactoring conditional directives for incomplete if conditions.

wmarquesr:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

* [Linux kernel coding style](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892)
* [Living in the #ifdef Hell](https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/)

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/kmod/igb/igb_main.c
+++ b/kmod/igb/igb_main.c
@@ -25,12 +25,12 @@
 /*
  * NOTE: This is an AVB-customed version of the standard Intel igb
  * driver. This driver requires (4) tx-rx queues are enabled, with MSI-X.
- * Standard best-effort LAN traffic is directed to tx queue(3), leaving tx-queue(0) 
- * through tx-queue(2) on the Intel i210 available to be mapped into a 
+ * Standard best-effort LAN traffic is directed to tx queue(3), leaving tx-queue(0)
+ * through tx-queue(2) on the Intel i210 available to be mapped into a
  * user-space application and managed by the application.
  *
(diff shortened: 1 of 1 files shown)


andrew-elder:
Thank you for your pull request. Do you  mind redoing the changes against the open-avb-next branch and doing a pull request to that branch? All new code and improvements go in to the open-avb-next branch before being periodically merged to master. You points on coding style are good ones. At the same time, I don't think we should generate a HUGE pull request with lots of formatting changes only. That doesn't help when reviewing change logs. But, if there are corrections that improve readability of the code, then I'm not opposed.

christopher-s-hall:
I agree with everything Andrew said and would like to add that, in the interest of future maintainability, we should not make a lot of changes that are not also in the kernel igb driver.  Maybe it would be better to submit changes to the igb driver and then when they're accepted pull them in here.

andrew-elder:
Chris, thank you for your comments. Makes sense to me. I wonder if we are close to the point where Open-AVB no longer needs to contain the igb driver? That way it would be 100% obvious not to submit changes back to Open-AVB. If Open-AVB still requires igb driver, then we need to add to the readme on the front page directing folks to the correct venue that will accept patches.

andrew-elder:
Pull request on wrong branch, so closing.

PR #334, opened 2015-12-15
Fixes: Wrong output from log messages on 64bit machines.

MarcinMiklas:
Details: Used macros for print formating eg. PRIu64 or PRIu32 so the expected
         and supplied formats match. Also changed format specifiers when they weren't matching
         expected format: eg. "llu" to "lu".


andrew-elder:

christopher-s-hall:
Looks OK to me too

Issue #338, opened 2016-01-08
MRP: listener merge call can cause incorrect listener declaration


andrew-elder:
The following sequence has been observed
1.  Listener end station client declares **Listener Asking Failed** (because Talker declaration has not yet been observed)
2.  Listener end station observes Talker declaration and then declares **Listener Ready**. Listener's mrpd waits for a transmit opportunity
3. Switch sends **rIn!** to end station declaring **Listener Asking Failed**. Merge operation in https://github.com/AVnu/Open-AVB/blob/master/daemons/mrpd/msrp.c#L276 causes declaration to be changed from the desired **Listener Ready** to **Listener Asking Failed**.

Captured logs for 5 streams follow (state 1 is Listener Asking Failed and state 2 is Listener Ready)

Listener declares **Listener Asking Failed**

	Jan  1 00:00:12 mrp: MRPD 012.013582 MSRP Listener (state = 1) event New!, VO->VN/MT
	Jan  1 00:00:12 mrp: MRPD 012.014913 MSRP Listener (state = 1) event New!, VO->VN/MT
	Jan  1 00:00:12 mrp: MRPD 012.016261 MSRP Listener (state = 1) event New!, VO->VN/MT
	Jan  1 00:00:12 mrp: MRPD 012.039990 MSRP Listener (state = 1) event New!, VO->VN/MT
	Jan  1 00:00:12 mrp: MRPD 012.043237 MSRP Listener (state = 1) event New!, VO->VN/MT
	Jan  1 00:00:12 mrp: MRPD 012.113639 MSRP Listener (state = 1) event tx!, VN->AN/MT
	Jan  1 00:00:12 mrp: MRPD 012.113737 MSRP Listener (state = 1) event tx!, VN->AN/MT
	Jan  1 00:00:12 mrp: MRPD 012.113766 MSRP Listener (state = 1) event tx!, VN->AN/MT
	Jan  1 00:00:12 mrp: MRPD 012.113790 MSRP Listener (state = 1) event tx!, VN->AN/MT
	Jan  1 00:00:12 mrp: MRPD 012.113814 MSRP Listener (state = 1) event tx!, VN->AN/MT
	Jan  1 00:00:12 mrp: MRPD 012.113944 MSRP send PDU

Listener declares **Listener Ready**

	Jan  1 00:00:12 mrp: MRPD 012.213363 MSRP Listener (state = 2) event New!, AN/MT
[...]

PR #339, opened 2016-01-08
MRP: correct listener attribute merge bug

andrew-elder:
This set of checkins restricts listener attribute merging to only occur in the direction of listener attribute propagation. It address issue #338 where a received PDU from a switch overwrote the declared listener attribute.

### DIFF
--- a/daemons/mrpd/msrp.c
+++ b/daemons/mrpd/msrp.c
@@ -99,6 +99,7 @@ void msrp_print_debug_info(int evt, const struct msrp_attribute *attrib)
 
 		}
 #endif
+#if 0
 		if (MSRP_TALKER_ADV_TYPE == attrib->type) {
 			state_mc_states = mrp_print_status(
 						&(attrib->applicant),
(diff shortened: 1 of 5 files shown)


davidcemin:
It all looks good to me. I will cross compile and put in our setup to verify that everything continues working properly.

Thank you Andrew

andrew-elder:
I'm in the process of renaming things for clarity, so this pull request will be updated sometime this week.

andrew-elder:
@davidcemin - if you could run this one, that would be great!

eric-mann:
... question about the merging of TA failure codes ... if a switch sends back to the talker failure code information on a talker advertise attribute, it doesn't appear that the failure code will be propagated back into the attribute (?).  Is this the desired behavior?

andrew-elder:
@eric-mann - you are correct, regarding declared TA being changed to TF by switch and reflected back, the failure code would not get propagated back in to the declared TA (in this modified code). I've pinged some experts to figure out if this is desirable or not. Streaming would not start anyway because Listener attribute would be Ask Failed. And you know whether or not you are asCapable, so it should not be an issue. I won't merge any of this until all interested parties agree.

pinealservo:
MRP propagation doesn't work backward.  Talker attributes are not supposed to get echoed back to the talker that declares them. Any notification that their declarations did not reach listeners successfully must come out-of-band or via Listener failure declarations.

andrew-elder:
Thanks for the input @pinealservo.

andrew-elder:
Testing has been completed, so merging this.

PR #341, opened 2016-01-14
igb_avb: Support 32bit app on 64bit kernel.

JaehyungAhn:
I got a problem that 32-bit application(talker) fails to run on 64-bit kernel.
There are two problems.
* 'compat_ioctl' callback should be added on file_operations.
* If memory address for DMA is over 32bits, user application calls 'mmap()' with wrong address.<BR/>
 In detail,<BR/>
 Application queries memory address for 'mmap()' with ioctl().<BR/>
 But if the address for DMA is reserved bigger than 4GB(32bits),<BR/>
 application gets clipped address on the result of ioctl().<BR/>
 After then, wrong address is used at calling 'mmap()'.

And one more trivial patch is added for typo.

### DIFF
--- a/kmod/igb/igb_main.c
+++ b/kmod/igb/igb_main.c
@@ -2602,7 +2602,7 @@ static int igb_probe(struct pci_dev *pdev,
 			pci_using_dac = 1;
 	} else {
 		err = dma_set_mask(pci_dev_to_dev(pdev), DMA_BIT_MASK(32));
-		if (err) {
+		if (!err) {
 			err = dma_set_coherent_mask(pci_dev_to_dev(pdev),
 				DMA_BIT_MASK(32));
 			if (err) {
--- a/kmod/igb/igb_main.c
+++ b/kmod/igb/igb_main.c
@@ -287,6 +287,9 @@ static struct file_operations igb_fops = {
 		.release = igb_close_file, 
 		.mmap	= igb_mmap, 
 		.unlocked_ioctl = igb_ioctl_file,
+#if defined(CONFIG_IGB_SUPPORT_32BIT_IOCTL)
+		.compat_ioctl = igb_ioctl_file,
+#endif
 };
 
 static struct miscdevice igb_miscdev = {
@@ -2594,6 +2597,7 @@ static int igb_probe(struct pci_dev *pdev,
 		return err;
 
 	pci_using_dac = 0;
+#if !defined(CONFIG_IGB_SUPPORT_32BIT_IOCTL)
 	err = dma_set_mask(pci_dev_to_dev(pdev), DMA_BIT_MASK(64));
 	if (!err) {
 		err = dma_set_coherent_mask(pci_dev_to_dev(pdev),
@@ -2601,6 +2605,7 @@ static int igb_probe(struct pci_dev *pdev,
 		if (!err)
 			pci_using_dac = 1;
 	} else {
+#endif
 		err = dma_set_mask(pci_dev_to_dev(pdev), DMA_BIT_MASK(32));
 		if (!err) {
 			err = dma_set_coherent_mask(pci_dev_to_dev(pdev),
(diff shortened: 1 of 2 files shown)


JaehyungAhn:
Sorry, one patch is updated just now. - 'igb_avb : support 32-bit..."
GFP_DMA32 is available according to defconfig.

andrew-elder:
Looks good to me. Anyone at Intel?

JaehyungAhn:
Exactly, I'm working for Windriver in Korea.
Windriver is Intel's Subsidiary company.
I had a chance to support eAVB driver recently. :)

eric-mann:
I'm looking at it as well ... :)

andrew-elder:
@eric-mann - I assume this pull request looks good to you?

eric-mann:
This patch only covers the case of transmit data buffer allocation (the "4K" pages) - wouldn't the transmit descriptor ring addresses have similar problems? Or are you assuming the kernel is 32-bit as well (?).

JaehyungAhn:
This patch also covers ring buffer addresses for transmission.
'dma_set_mask()' is called on igb_probe, and my patch restricts memory address for dma to 32-bit.
So, if device for igb_avb requests dma buffers, the request buffers are not beyond 4GB(32 bits address space).

eric-mann:
32-bit mode has been generally 'hard' for us to support for our standard drivers, hence why I'm scrutinizing this patch. 

What I don't understand is the original post - "32-bit application(talker) fails to run on 32-bit kernel.". The changes seem to have to desired effect to limit the DMA addresses to be < 4GB (32-bit) - but if the kernel is 32-bit why would the kernel ever return more than 32-bit addresses?

It would make more sense if you stated "32-bit application(talker) fails to run on 64-bit kernel."

What I'm also trying to figure out is if any of the mapped regions are missing or not covered - the driver returns 
(a) DMA buffers for packets (which you covered), 
(b) descriptor ring buffers which you state is covered by the dma_set_mask() in igb_probe(), and
(c) the device registers mapped into user-space. 

The device has 64-bit BARs ... do you think there are any cases where the physical memory allocation of the BARs themselves will cause a problem?

JaehyungAhn:
Ah... Sorry. I had mistaken on my first message.
Your guess is right, as the title of the 'Pull Request'
This issue is for '32-bit application(talker) fails to run on 64-bit kernel.'
(I have just fixed the message)

And, usually... PCI BARs are reserved below 4G, so I didn't get a problem for PCI BARs.
But, not sure and BIOS/UEFI has the role for setting address/values of BARs as I know.
( refer -  http://comments.gmane.org/gmane.linux.kernel.kernelnewbies/27637 )
So, I think, PCI BARs should be reserved below 4G by BIOS/UEFI to make sure 32-bit compatibility.(even though generally it does...)

At last... I have also checked if there are *not covered* cases.
There are 3 cases to use mmap.
1. PCI BAR - pci mmio area - to control eAVB on user application.
2. ring buffer - seems to be used Tx/Rx and allocated with dma_alloc_coherent().
3. map buffer - also seems to be used Tx/Rx but allocated with alloc_page().
                     So, in this case, I add flag - GFP_DMA(32).
                     If GFP_DMA(32) is added, 4K page is taken from DMA_ZONE, so address would be below 4G.

eric-mann:
Looks like remap_pfn_range() will do the right thing and map the 64-bit physical address into the 32-bit process address space... so we can go ahead and accept the pull request.

andrew-elder:
Thanks Eric. I'll go ahead and merge.

JaehyungAhn:
There is a wrong description on my note.
(But, no effect to patches.)
I said, PCI BARs should be reserved below 4G, but it's not.
In the case of mapping PCI resource, 'offset' parameter is '0'.
So, driver can map PCI IO areas to application's virtual address.
The problem which my patch handles is the case that the 'offset' parameter is not '0'.(queried address by ioctl)

Issue #134, opened 2014-11-07
gPTP.com.c.9.2,  timeBaseIndicator changes at wrong time


andrew-elder:
gPTP.com.c.9.2, should only change timeBaseIndicator when GM phase or freq changes

PR #342, opened 2016-01-25
Issue #134 fix

davidcemin:
This patch fixes Issue #134 

The clause 9.2 from 802.1AS-2011 states that the "time base indicator should change its value if and only if there is a phase or frequency change". 

This patch addresses this problem by incrementing the TBI every time there is a change of ptp master. Also, the values lastGmPhaseChange and lastGmFreqChange were not being updated correctly.

To do that  i did:
- Created two FollowUpTLV instances inside the IEEE15588Clock class; One of them is updated every time the FollowUp::processMessage method is called (fup_status) and the second one update its value everytime the TBI needs to be changed (fup_info) with the instantaneous content of fup_status. 
- Before sending out our FUP messages, the function that prepares the packet gets the information from fup_status and put inside the packet.

This issue was validated using the IOL Tools on the specific tests that exercises this part of the standard.


daemons/gptp/common/avbts_clock.hpp
```diff
@@ -130,7 +130,21 @@ class IEEE1588Clock {
...
+     * the base time has changed. When that happens
+     * the information from fup_status is copied over
+     * fup_info. The follow-up sendPort method is
+     * supposed to get the information from fup_info
+     * prior to sending a message out.
+     */
+    FollowUpTLV *fup_info;
+
+    /**
+     * fup status has the instantaneous info*/
```
andrew-elder: I would have expected the */ to be on a new line.

daemons/gptp/common/avbts_message.hpp
```diff
@@ -622,7 +655,7 @@ class FollowUpTLV {
 	int32_t cumulativeScaledRateOffset;
 	uint16_t gmTimeBaseIndicator;
 	scaledNs scaledLastGmPhaseChange;
-	int32_t scaledLastGmFreqChange;
+    int32_t scaledLastGmFreqChange;
```
andrew-elder: Looks like different indent from the preceding line.

daemons/gptp/common/avbts_message.hpp
```diff
@@ -655,6 +688,72 @@ class FollowUpTLV {
...
+
+    /**
+     * @brief  Sets the gmTimeBaseIndicator private member
+     * @param  tbi Value to be set
+     * @return void
+     */
+    void setGMTimeBaseIndicator(uint16_t tbi)
+    {
+        gmTimeBaseIndicator = PLAT_htonl(tbi);
+    }
```
andrew-elder: This implementation conflicts with the inc method below. ie can't increment a multi-byte variable using ++ if it is stored in network order. I'm actually opposed to storing anything in network order, but that may be too large of a change at this point - and out of scope for this work.

daemons/gptp/common/ieee1588port.cpp
```diff
@@ -863,7 +864,8 @@ void IEEE1588Port::processEvent(Event e)
 			 (pow ((double)2, getAnnounceInterval()) * 1000000000.0));
 		break;
 	case FAULT_DETECTED:
-		XPTPD_INFO("Received FAULT_DETECTED event");
+        XPTPD_INFO("Received FAULT_DETECTED event");
+        setAsCapable(false);
```
andrew-elder: check indenting

daemons/gptp/common/ieee1588port.cpp
```diff
@@ -941,7 +945,7 @@ void IEEE1588Port::becomeSlave( bool restart_syntonization ) {
   clock->deleteEventTimer( this, ANNOUNCE_INTERVAL_TIMEOUT_EXPIRES );
   clock->deleteEventTimer( this, SYNC_INTERVAL_TIMEOUT_EXPIRES );
 
-  port_state = PTP_SLAVE;
+    port_state = PTP_SLAVE;
```
andrew-elder: check indenting

daemons/gptp/common/ieee1588port.cpp
```diff
@@ -980,6 +986,7 @@ void IEEE1588Port::recommendState
 		  if( changed_external_master ) {
 		    XPTPD_PRINTF("Changed master!\n" );
 		    clock->newSyntonizationSetPoint();
+            getClock()->updateFUPInfo();
```
andrew-elder: check indenting

daemons/gptp/common/ptp_message.cpp
```diff
@@ -873,7 +875,7 @@ void PTPMessageFollowUp::sendPort(IEEE1588Port * port,
 	XPTPD_INFO("Follow-up Dump:\n");
 #ifdef DEBUG
 	for (int i = 0; i < messageLength; ++i) {
-		XPTPD_PRINTF("%d:%02x ", i, (unsigned char)buf[i]);
+		fprintf(stderr, "%d:%02x ", i, (unsigned char)buf_t[i]);
```
andrew-elder: Is there a reason to remove the macro implementation? With a macro others can override implementation more easily.

daemons/gptp/common/ptp_message.cpp
```diff
@@ -898,6 +900,8 @@ void PTPMessageFollowUp::processMessage(IEEE1588Port * port)
 	FrequencyRatio local_system_freq_offset;
 	FrequencyRatio master_local_freq_offset;
 	int correction;
+    int32_t scaledLastGmFreqChange = 0;
+    scaledNs scaledLastGmPhaseChange;
```
andrew-elder: check indent

daemons/gptp/common/ptp_message.cpp
```diff
@@ -983,6 +987,12 @@ void PTPMessageFollowUp::processMessage(IEEE1588Port * port)
 	  calcMasterLocalClockRateDifference
 	  ( preciseOriginTimestamp, sync_arrival );
 
+    /*Update information on local status structure.*/
+    scaledLastGmFreqChange = (int32_t)((1.0/local_clock_adjustment -1.0) * (1ULL << 41));
+    scaledLastGmPhaseChange.setLSB( tlv.getRateOffset() );
+    port->getClock()->getFUPStatus()->setScaledLastGmFreqChange( scaledLastGmFreqChange );
+    port->getClock()->getFUPStatus()->setScaledLastGmPhaseChange( scaledLastGmPhaseChange );
+
```
andrew-elder: check indent

daemons/gptp/gptp_cfg.ini
```diff
@@ -18,19 +20,23 @@ syncReceiptTimeout = 3
 # various links, in clause 11.2.2. It needs to be configurable because a
 # user may be using a converter to fibre and the software wont know until the
 # administrator changes it.
-neighborPropDelayThresh = 800
+neighborPropDelayThresh = 8000
```
andrew-elder: what is the motivation for making this so large here?

andrew-elder:
I'm done with my comments. Mostly looks good apart from a few places to check indenting. Only really issue was the ++ operation on a variable stored in network order.

daemons/gptp/gptp_cfg.ini
```diff
@@ -1,7 +1,9 @@
 
 [ptp]
 
-#
+# Priority1 value
+# It can assume values between 0 and 255.
+# The lower the number, the higher the priority for the BCMA.
```
pinealservo: BCMA should be BMCA

daemons/gptp/linux/src/daemon_cl.cpp
```diff
@@ -325,10 +319,26 @@ int main(int argc, char **argv)
             fprintf(stdout, "phy_delay_mb_tx: %d\n", iniParser.getPhyDelayMbTx());
             fprintf(stdout, "phy_delay_mb_rx: %d\n", iniParser.getPhyDelayMbRx());
             fprintf(stdout, "neighborPropDelayThresh: %ld\n", iniParser.getNeighborPropDelayThresh());
+
+            /* If using config file, set the neighborPropDelayThresh.
+             * Otherwise it will use its default value (800ns) */
             port->setNeighPropDelayThresh(iniParser.getNeighborPropDelayThresh());
+
+            /*Only overrites phy_delay default values if not input_delay switch enabled*/
```
pinealservo: overrites should be overwrites

pinealservo:
I just found a couple of minor things in comments; looks good to me.

davidcemin:
Thanks ! I will address all the comments before moving to the next Issue. (i.e. probably by tomorrow it will be done)

daemons/gptp/common/avbts_clock.hpp
davidcemin: ok

daemons/gptp/common/avbts_message.hpp
davidcemin: Yes, my editor defaults to spaces instead of tabs and I apparently touched this line. Im putting a tab here instead.

daemons/gptp/common/avbts_message.hpp
davidcemin: You are right. I just kept the same format of the constructor of this method, which initializes it with "0" on the network order. 
since we are already doing it on network order, I think I should keep this method the way it is, and below I increment it on host order and then store it back on network order. 
Would that be acceptable for now ? We can open another issue to investigate storing this on network order vs host order so we dont forget about this as well.

andrew-elder: Yes, that would be fine.

davidcemin: The other option would be storing it properly on the host order, and converting to the network order on the FollowUP::sendPort method

andrew-elder: If you are sure there are no side effects, I would favor late conversion to network order in the ::sendPort method.

davidcemin: Currently its not being used anywhere else, and I prefer the latter as well.

christopher-s-hall: I think this (Andrew's suggestion) is the convention that we decided on a while ago. We should be consistent.

daemons/gptp/common/ptp_message.cpp
```diff
@@ -933,7 +937,7 @@ void PTPMessageFollowUp::processMessage(IEEE1588Port * port)
 	sync_arrival = sync->getTimestamp();
 
 	delay = port->getLinkDelay();
-	if ((delay = port->getLinkDelay()) == 3600000000000) {
+	if ((delay = port->getLinkDelay()) == 3600000000000) { /*TODO: ONE_WAY_DELAY_DEFAULT*/
```
christopher-s-hall: There is a magic number problem here (of my creation). This is not meant to be a "default" value it is supposed to be an incorrect value that should not be used (aborting processing).  Maybe the code should be something like:

if(( delay = port->getLinkDelay()) == INVALID_LINKDELAY ) {

daemons/gptp/common/ieee1588port.cpp
davidcemin: Ok, same problem as above. I have spaces instead of tabs on my vim. I rolled it back and kept these two lines as tabs.

daemons/gptp/common/ieee1588port.cpp
davidcemin: This file has 3 different indenting formats: 2 spaces, 4 spaces and tabs. For this particular line, it is two spaces, so I kept it like that.

daemons/gptp/common/ieee1588port.cpp
davidcemin: and here its tab on the first indent and 4 spaces on the second.

daemons/gptp/common/ptp_message.cpp
davidcemin: No, i've changed it to printf to debug and forgot to change it back. Im keeping the XPTPD_PRINTF macro

daemons/gptp/gptp_cfg.ini
davidcemin: I did that on october, and I think I was having trouble to synchronize it against one Extreme switch. I will keep the default to 800 though.

daemons/gptp/linux/src/daemon_cl.cpp
davidcemin: Thanks. Two typos :)

daemons/gptp/common/ptp_message.cpp
davidcemin: I can change that now since Im touching this code. So, INVALID_LINKDELAY should be the 3600000000000 ?

christopher-s-hall: That would make this more clear. The other possibility would be to add a boolean such as link_delay_valid or something like that.

bool link_delay_is_valid = port->getLinkDelay( &delay );
if( !link_delay_valid ) {
  goto done;
}

// Else continue processing message

The port code may already have a link delay valid flag.

andrew-elder: Shouldn't the if statement read
if (delay == 3600000000000)
?
previous line before the if() sets the value of delay.

christopher-s-hall: Yes. It's probably somewhat confusing for the if conditional to have the side effect of setting the value of delay.

davidcemin: So I guess this would do:

-       delay = port->getLinkDelay();
-       if ((delay = port->getLinkDelay()) == 3600000000000) { /*TODO: ONE_WAY_DELAY_DEFAULT*/
+       if( !port->getLinkDelay(&delay) ) {
                goto done;

daemons/gptp/common/ptp_message.cpp
```diff
@@ -932,8 +940,7 @@ void PTPMessageFollowUp::processMessage(IEEE1588Port * port)
 
 	sync_arrival = sync->getTimestamp();
 
-	delay = port->getLinkDelay();
-	if ((delay = port->getLinkDelay()) == 3600000000000) {
+	if( !port->getLinkDelay(&delay) ) {
 		goto done;
```
andrew-elder: This looks ok to me.

daemons/gptp/common/ptp_message.cpp
```diff
@@ -983,6 +990,12 @@ void PTPMessageFollowUp::processMessage(IEEE1588Port * port)
 	  calcMasterLocalClockRateDifference
 	  ( preciseOriginTimestamp, sync_arrival );
 
+    /*Update information on local status structure.*/
+	scaledLastGmFreqChange = (int32_t)((1.0/local_clock_adjustment -1.0) * (1ULL << 41));
```
andrew-elder: Indent of comment!

Issue #135, opened 2014-11-07
gPTP.com.c.13.2, FollowUp sequence ID issue


andrew-elder:
gPTP.com.c.13.2, check FollowUp sequence IDs

davidcemin:
The proposed solution to this issue is:
https://github.com/AVnu/Open-AVB/blob/feature-gptp-next/daemons/gptp/common/ptp_message.cpp#L934

This verification just tests if they are different and return. The idea is to add an error counter inside this if, and a configurable threshold (through the .ini file) to compare it against. If the counter crosses the threshold, then we call port->becomeMaster(true), which will force an announce to be called.

Does anyone see a problem on this approach ?

andrew-elder:
Seems ok to me, but I'm not an expert. @christopher-s-hall - what do you think?

davidcemin:
I implemented the idea explained on the comment above and I created the pull request already. We can discuss if this is valid or not and I will change it if necessary.

PR #343, opened 2016-01-28
Fix wrong behaviour with wrong seqIDs on sync messages 

davidcemin:
Fixes Issue #135 

On the followUP messages, it is checked if the amount of wrong syncMessages is over a threshold, and if that is true, gPTP switches to master.

This patch was validated using the UNH-IOL tools for gPTP.

### DIFF
--- a/daemons/gptp/common/avbts_message.hpp
+++ b/daemons/gptp/common/avbts_message.hpp
@@ -806,6 +806,11 @@ public:
 		preciseOriginTimestamp = timestamp;
 	}
 
+	/**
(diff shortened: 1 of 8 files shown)


andrew-elder:
This change set looks good to me.

davidcemin:
does anyone else had the time to take a look at this patch ? It would be great to have it merged to the feature-gptp-next branch before i create another pull request for next issue.

andrew-elder:
@davidcemin - if I don't see any further comments, I'll merge this pull request first thing in the morning.

Issue #137, opened 2014-11-07
gPTP.com.c.15.3:B, handle lost pDelay responses


andrew-elder:
gPTP.com.c.15.3:B, Lose more than 3 consecutive Pdelay_Resp_Follow_Up messages.

Issue #136, opened 2014-11-07
gPTP.com.c.15.2:C, incorrect sequenceID


andrew-elder:
gPTP.com.c.15.2:C, incorrect sequenceId in response afters asCapable

PR #344, opened 2016-02-03
Adding threshold for different parameters within gptp

davidcemin:
This patch fixes both Issues #136 and #137.

In both cases, a new configuration item was added to the .ini file.

### DIFF
--- a/daemons/gptp/common/avbts_port.hpp
+++ b/daemons/gptp/common/avbts_port.hpp
@@ -234,6 +234,8 @@ class IEEE1588Port {
 	static const int64_t INVALID_LINKDELAY = 3600000000000;
 	static const int64_t NEIGHBOR_PROP_DELAY_THRESH = 800;
 	static const unsigned int SYNC_RECEIPT_THRESH = 5;
+	static const unsigned int SEQID_ASCAPABLE_THRESHOLD = 5;
+
 	/* Signed value allows this to be negative result because of inaccurate
 	   timestamp */
 	int64_t one_way_delay;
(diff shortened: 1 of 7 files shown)


daemons/gptp/common/avbts_port.hpp
```diff
@@ -234,6 +234,9 @@ class IEEE1588Port {
 	static const int64_t INVALID_LINKDELAY = 3600000000000;
 	static const int64_t NEIGHBOR_PROP_DELAY_THRESH = 800;
 	static const unsigned int SYNC_RECEIPT_THRESH = 5;
+	static const unsigned int SEQID_ASCAPABLE_THRESHOLD = 5;
+	static const uint16_t LOSTPDELAY_RESP_THRESH = 3;
```
andrew-elder: Why aren't these called DEFAULT_SEQID_ASCAPABLE_THRESHOLD etc.?
They are really just initial values, correct?
They are overridden by the .ini file - at least that is what I assume.

daemons/gptp/common/avbts_port.hpp
```diff
@@ -243,6 +246,14 @@ class IEEE1588Port {
 	unsigned int sync_receipt_thresh;
 	unsigned int wrongSeqIDCounter;
 
+    /*SeqID threshold*/
```
andrew-elder: indent

daemons/gptp/common/ptp_message.cpp
```diff
@@ -1359,32 +1360,50 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
 	req->getPortIdentity(&req_id);
 	resp->getRequestingPortIdentity(&resp_id);
 
-	if (req->getSequenceId() != sequenceId) {
+    /*We need to keep track of lost respFUPs. If we have more
```
andrew-elder: check indent

andrew-elder:
Overall looks good to me.

daemons/gptp/common/avbts_port.hpp
davidcemin: Yes, I can change the name and they are overridden by the .ini file if a .ini file is used.

daemons/gptp/common/ptp_message.cpp
davidcemin: This file mostly uses tabs, so i will keep tabs instead of spaces.

Issue #138, opened 2014-11-07
gPTP.com.c.15.3:C, handle mix of list resp and followups.


andrew-elder:
gPTP.com.c.15.3:C, Consecutively lose a mix of Pdelay_Resp and Pdelay_Resp_Follow_Up messages.

PR #349, opened 2016-02-11
Handles lost of Resp and Fup messages

davidcemin:
This patch fixes Issue #138 and potentially others as well. There are other open issues that are related to handle fups/resp lost messages and making gPTP not being asCapable. The Issue #138 was fixed and I am retesting the whole thing to see if at this point nothing got broken and if there are other ones fixed as well. 

I will be posting the other results directly on the open issues if I have good news. For now I want to create the pull request (which is quite small - but required lot of investigation and testing) so people can review and we can move forward and integrate this into Avnu's repo.

### DIFF
--- a/daemons/gptp/common/avbts_message.hpp
+++ b/daemons/gptp/common/avbts_message.hpp
@@ -949,6 +949,8 @@ class PTPMessagePathDelayRespFollowUp:public PTPMessageCommon {
 	PortIdentity *requestingPortIdentity;
 
 	PTPMessagePathDelayRespFollowUp(void) { }
+
+	const static uint16_t SEQID_DIFF_THRESHOLD = 4;
 public:
 	/**
 	 * Builds the PTPMessagePathDelayRespFollowUp object
--- a/daemons/gptp/common/avbts_message.hpp
+++ b/daemons/gptp/common/avbts_message.hpp
@@ -949,8 +949,6 @@ class PTPMessagePathDelayRespFollowUp:public PTPMessageCommon {
 	PortIdentity *requestingPortIdentity;
 
 	PTPMessagePathDelayRespFollowUp(void) { }
-
-	static const unsigned int SEQID_DIFF_THRESHOLD = 5;
 public:
 	/**
 	 * Builds the PTPMessagePathDelayRespFollowUp object
--- a/daemons/gptp/common/avbts_message.hpp
+++ b/daemons/gptp/common/avbts_message.hpp
@@ -950,7 +950,6 @@ class PTPMessagePathDelayRespFollowUp:public PTPMessageCommon {
 
 	PTPMessagePathDelayRespFollowUp(void) { }
 
-	const static uint16_t SEQID_DIFF_THRESHOLD = 4;
 public:
 	/**
 	 * Builds the PTPMessagePathDelayRespFollowUp object
--- a/daemons/gptp/common/avbts_message.hpp
+++ b/daemons/gptp/common/avbts_message.hpp
@@ -949,6 +949,8 @@ class PTPMessagePathDelayRespFollowUp:public PTPMessageCommon {
 	PortIdentity *requestingPortIdentity;
 
 	PTPMessagePathDelayRespFollowUp(void) { }
+
+	static const unsigned int SEQID_DIFF_THRESHOLD = 5;
 public:
 	/**
 	 * Builds the PTPMessagePathDelayRespFollowUp object
--- a/daemons/gptp/common/ptp_message.cpp
+++ b/daemons/gptp/common/ptp_message.cpp
@@ -1323,6 +1323,7 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
 	Timestamp request_tx_timestamp(0, 0, 0);
 	Timestamp remote_req_rx_timestamp(0, 0, 0);
 	Timestamp response_rx_timestamp(0, 0, 0);
+	unsigned int diffLastSeqId;
 
 	if (port->getPortState() == PTP_DISABLED) {
 		// Do nothing all messages should be ignored when in this state
@@ -1354,6 +1355,7 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
 		/* Probably shouldn't happen either */
 		XPTPD_ERROR
 		    (">>> Received PDelay followup but no RESPONSE exists");
+
 		goto abort;
 	}
 
@@ -1364,17 +1366,18 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
 	 * than a certain number of lost FUPs we should set asCapable
 	 * to false
 	 */
+	diffLastSeqId = sequenceId - port->getLastSeqId();
 	if( sequenceId != (port->getLastSeqId() + 1) )
 	{
 		XPTPD_ERROR("Current seqID: %d. %d are missing since the last one sent",
 				sequenceId, sequenceId-port->getLastSeqId());
-		if( (sequenceId - port->getLastSeqId() ) > port->getLostPdelayRespThresh() )
+		if( diffLastSeqId > port->getLostPdelayRespThresh() )
 		{
+			XPTPD_ERROR(">>> Number of missing pDelayResponse messages is bigger than threshold.");
 			port->setAsCapable( false );
 		}
 	}
 	port->setLastSeqID(sequenceId);
-
 
 	// Check if we have received a response
 	/* Count wrong seqIDs and after a threshold (from .ini), mark
@@ -1386,8 +1389,11 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
 		uint16_t req_port_number;
 		resp_id.getPortNumber(&resp_port_number);
 		requestingPortIdentity->getPortNumber(&req_port_number);
-		if( !port->incSeqIdAsCapableThreshCounter() )
+		bool isSeqIdCounterUnderThresh = port->incSeqIdAsCapableThreshCounter();
+		if( !isSeqIdCounterUnderThresh ||
+				(port->getSeqIdAsCapableThreshCounter() + diffLastSeqId) > SEQID_DIFF_THRESHOLD)
 		{
+			XPTPD_ERROR(">>> SeqID counter bigger than threshold.");
 			port->setAsCapable( false );
 		}
 		XPTPD_ERROR
@@ -1400,10 +1406,6 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
 
 		goto abort;
 	}
-    else
-    {
-        port->setSeqIdAsCapableThreshCounter(0);
-    }
 
 	XPTPD_INFO("Request Sequence Id: %u", req->getSequenceId());
 	XPTPD_INFO("Response Sequence Id: %u", resp->getSequenceId());
@@ -1467,6 +1469,8 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
 		 remote_req_rx_timestamp.nanoseconds);
 
 	// Adjust turn-around time for peer to local clock rate difference
+	// TODO: Are these .998 and 1.002 specifically defined in the standard?
+	// Should we create a define for them ?
 	if
 		( port->getPeerRateOffset() > .998 &&
 		  port->getPeerRateOffset() < 1.002 ) {
--- a/daemons/gptp/common/ptp_message.cpp
+++ b/daemons/gptp/common/ptp_message.cpp
@@ -1389,11 +1389,13 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
 		uint16_t req_port_number;
 		resp_id.getPortNumber(&resp_port_number);
 		requestingPortIdentity->getPortNumber(&req_port_number);
+		uint16_t _cntthres = port->getSeqIdAsCapableThreshCounter() + diffLastSeqId;
 		bool isSeqIdCounterUnderThresh = port->incSeqIdAsCapableThreshCounter();
-		if( !isSeqIdCounterUnderThresh ||
-				(port->getSeqIdAsCapableThreshCounter() + diffLastSeqId) > SEQID_DIFF_THRESHOLD)
+		bool cntOverThresh = _cntthres > port->getSeqIdAsCapableThresh();
+
+		if( !isSeqIdCounterUnderThresh || cntOverThresh )
 		{
-			XPTPD_ERROR(">>> SeqID counter bigger than threshold.");
+			XPTPD_ERROR(">>> SeqID counter bigger than threshold.(%u > %u)", _cntthres, port->getSeqIdAsCapableThresh());
 			port->setAsCapable( false );
 		}
 		XPTPD_ERROR
--- a/daemons/gptp/common/ptp_message.cpp
+++ b/daemons/gptp/common/ptp_message.cpp
@@ -1361,7 +1361,7 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
 
 	if( req->getSequenceId() != sequenceId ) {
 		XPTPD_ERROR
-			(">>> Received PDelay FUP has different seqID than the PDleay request (%d/%d)",
+			(">>> Received PDelay FUP has different seqID than the PDelay request (%d/%d)",
 			 sequenceId, req->getSequenceId() );
 		goto abort;
 	}
--- a/daemons/gptp/common/ptp_message.cpp
+++ b/daemons/gptp/common/ptp_message.cpp
@@ -1228,8 +1228,6 @@ void PTPMessagePathDelayResp::processMessage(IEEE1588Port * port)
 		return;
 	}
 
-	port->getClock()->deleteEventTimerLocked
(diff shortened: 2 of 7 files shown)


daemons/gptp/common/ptp_message.cpp
```diff
@@ -1467,6 +1469,8 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
 		 remote_req_rx_timestamp.nanoseconds);
 
 	// Adjust turn-around time for peer to local clock rate difference
+	// TODO: Are these .998 and 1.002 specifically defined in the standard?
```
pinealservo: I couldn't find anything relating to a threshold for whether to do a rate ratio adjustment of turnaround time. According to 11.2.15.2.4 of 802.1AS, we should always scale `link_delay` by the neighbor rate ratio rather than `turn_around`, at least if I'm understanding the code and spec correctly.

Regardless of whether we're scaling the right term, I think that scaling only when the rate ratio is outside a range is likely to introduce an exaggerated step-change when the range boundary is crossed. It would probably be good to find the author of the code and ask what it's meant to do, though, if that's possible.

pinealservo:
Your changes look good to me.  My only concern was in existing code, so as far as I can tell this patch does no harm with respect to that.

daemons/gptp/common/ptp_message.cpp
```diff
@@ -1364,18 +1366,19 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
 	 * than a certain number of lost FUPs we should set asCapable
 	 * to false
 	 */
+	diffLastSeqId = sequenceId - port->getLastSeqId();
```
pinealservo: I have some concerns about this calculation, and possibly the comparison that follows.  I suspect it may run afoul of the integer promotion and usual arithmetic conversion rules.

pinealservo: Here's some sample code to illustrate what I mean:

```C
#include <stdint.h>
#include <stdio.h>

int main(void)
{
  uint16_t seq1 = UINT16_MAX;
  uint16_t seq2 = 0;
  uint16_t sD2;
  unsigned int seqDiff;

  seqDiff = seq2 - seq1;
  sD2 = seq2 - seq1;

  printf("seqDiff is %u, sd2 is %u\n", seqDiff, sD2);

  sD2 = seq1 + 1;
  printf("Tests are: %d %d\n", seq2 == (seq1 + 1), seq2 == sD2);

  return 0;
}
```

```
$ ./conversions 
seqDiff is 4294901761, sd2 is 1
Tests are: 0 1
```

pinealservo: Keep in mind that the "usual arithmetic conversions" apply to all the basic math operators. And the first step of those conversions, when dealing with integral operands, is to apply the "integer promotions".

The integer promotions will convert any integral type of *lesser rank* than the rank of `int` and `unsigned int` to be of type `int` so long as all the values of the original type will fit in `int`. Otherwise the type will be promoted to `unsigned int`.  Rank generally corresponds to the bit-width of the type, with wider types having higher rank, and the signed/unsigned versions having the same rank. So in this case, the `uint16_t` values are converted to `int`, since all `uint16_t` values fit in the range of `int`.

So the result of subtracting two `uint16_t` values is a value of type `int`.  In a wrap-around situation, it will be a small negative number.  Now, when that value is assigned to a variable, the value on the right hand side of the assignment expression takes on the type of the left-hand side.

In the case of the left-hand side having type `unsigned int`, the negative number can't be represented in that type. Therefore, the unsigned conversion rule applies, which repeatedly adds or subtracts one more than the maximum representable value of the target type to the value until the result falls in the representable range. In this example, we add `UINT_MAX+1` to `-1` and end up with `UINT_MAX`. Fortunately, C provides reasonable semantics for unsigned conversion and overflow; in the case of values that don't fit into signed types (or which overflow), the result is undefined.

andrew-elder: Good spot. So diffLastSeqId should be declared as type uint16_t, right?

andrew-elder: It would be nice to have a test framework to inject some of these wrap cases into the gPTP source code!

daemons/gptp/common/ptp_message.cpp
andrew-elder: My vote is that if this is not resolved today, this should not block pull request - we will create a gPTP github issue to review this at a later date.

daemons/gptp/common/ptp_message.cpp
davidcemin: Nice ! Thank you guys. I will try to get this done by today.

daemons/gptp/common/ptp_message.cpp
davidcemin: agreed. I'll open the issue by the end of the day (Eastern) if not yet clear.

daemons/gptp/common/ptp_message.cpp
davidcemin: So, as Andrew already mentioned, I understand by the code above and your (really nice) explanation about it that switching the unsigned int to a uint16_t would solve this problem. Is that correct? I will push in the diff already so if it is the case, we can move forward and integrate this

pinealservo:
As it stands, I think you'll get a spurious lost message log every time it wraps, but thanks to this fix it shouldn't cause the port's asCapable flag to turn off, so it's probably fine to do what you suggest.

daemons/gptp/common/ptp_message.cpp
christopher-s-hall: The numbers I put here are somewhat arbitrary. I have in the past used hardware that occasionally returned bad timestamps. Since it seemed like a difficult task to examine a single timestamp and decide whether it was good, I completed the calculation, and if, for example, one of the timestamps was bad, the result would be way off and the calculation would be rejected. There is probably a better way to accomplish this, but it did it's job at the time. I don't really have an opinion as whether to keep the code. It probably can go.

davidcemin:
I understand by your explanation that if I store the left part of the comparison (port->getSeqIdAsCapableThreshCounter() + diffLastSeqId) in a uint16_t variable, the problem of adding one in case of roll over should be avoided. Is that correct? 

Its important to notice that the counter should never pass the THRESHOLD value (which is set to 5), since once it gets to these value, the asCapable is set to False, and then it stops the increment. 

The other case of failure would be when the diffLastSeqId is a negative number, and in this case, the comparison will be set to true and hence we will have the gPTP not asCapable. The negative number will happen when port->lastSeqId() value is still set to UINT16_MAX (which should be when the current is set to a very small number, like 0 or 1). Do you see an easy way to avoid this besides adding more logic to calculate when it is rolling over?

andrew-elder:
See http://stackoverflow.com/questions/7221409/is-unsigned-integer-subtraction-defined-behavior. diffLastSeqId is never negative as it is unsigned. It could get "huge" if 32678 packets with sequence IDs are missed, that seems to be a failure condition in any case.

I didn't understand the first paragraph regarding "left part of the comparison (port->getSeqIdAsCapableThreshCounter() + diffLastSeqId) in a uint16_t variable". I would assume you are counting the number of times the seq ID does not increment by 1? Or, maybe if the diffLastSeqID is greater than some threshold? If the later, simple uint16_t subtraction and result storage will take care of rollover conditions.

davidcemin:
Ok, i will just post the code since its easier to explain what I mean.

davidcemin:
I think you are right, they might have the same value now, but we dont want the user changing both at the same time. I thought I could generalize that by removing the define. I will roll it back that part.

davidcemin:
Yes you are right about the resets.

pinealservo:
@andrew-elder Note that when you subtract two uint16_t variables, you are actually doing a *signed* subtraction because of the promotion rules. If they were of `unsigned int` or greater rank and unsigned, then there would be an unsigned subtraction. This makes little difference on a 2's complement machine if you convert the result back to an unsigned type, except for the fact that in C, signed overflow is undefined behavior while unsigned overflow is well-defined.

@davidcemin You can either store the result of the increment in a `uint16_t` variable or use a `uint16_t` cast to ensure you get the arithmetic modulo-2^16 behavior that you want.

daemons/gptp/common/ptp_message.cpp
```diff
@@ -1364,18 +1366,19 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
...
 	if( sequenceId != (port->getLastSeqId() + 1) )
```
pinealservo: This will fix the rollover error here, because the next sequence number after `UINT16_MAX` should be 0 rather than `(int)(UINT16_MAX+1)`, which it will be in the code as written due to integer promotion.

    if( sequenceId != (uint16_t)(port->getLastSeqId() + 1) )

pinealservo:
> There must be some convoluted conditional logic in 802.1AS if this is the code it produces! If diffLastSeqId is 1 repeatably code can execute many times before asCapable is reset.. If diffLastSeqId is 10 (larger than port->getSeqIdAsCapableThresh()) then asCapable is reset the first time though. Have I gotten that correct?

802.1AS just says that if you miss getting a certain number of PDelay responses (3), then the port is no longer asCapable and the tracking of lostResponses should reset. When you set asCapable to false as a result of lostResponses, there's a statistics counter `pdelayAllowedLostResponsesExceededCount` that should be incremented. 

Lines 1369-1380 attempt to do this via checking the sequentiality of the sequenceId values contained in the PathDelayRequestFollowUp messages. It's important to note that the `getLastSeqID` and `setLastSeqID` methods on the port are only used here; this does not refer to the sequenceId field of the `PathDelayRequest` message we sent out! If there is ever a gap that exceeds the threshold, asCapable is set to false. I don't think this is quite the correct thing to do (missed PathDelayResponse and PathDelayResponseFollowUp messages should be determined solely based on a timeout), but it does make some sense.

Lines 1386-1410 do a different sort of checking; they're checking for consistency between this PathDelayResponseFollowUp and the PathDelayResponse that we last received. I'm not sure what's going on with the code that deals with the failure of this consistency check, though. According to the state machines in the spec, we should just ignore messages with sequenceIds that don't match our request. If we fail to get a good one before the timeout occurs, the timeout handler should deal with the error condition.  For this to work, the PathDelayResponse handler would also have to check the sequenceId and discard if it doesn't match.

To sum up, I don't think we should have to do any of these "sequentiality checks" at all. The only sequenceId check that matters is the one for equality to the last request message we sent, which we're currently not doing at all!  It should be done in both the response and follow-up handlers, and non-matches or duplicate messages should count as a lost response. If we don't get a response by the time we are ready to send the next one, then we count that as a lost response too.  Whenever we have a lost response, if it's over the threshold we are no longer asCapable; otherwise we proceed to send a request message with the next sequenceId. When we *do* get a valid response and follow-up, we reset the lost response count. See Figure 11-8 in 802.1AS-2011 for the state machine.

andrew-elder:
@davidcemin - any comments on the above?

christopher-s-hall:
Lines 1386-1410 do a different sort of checking; they're checking for consistency between this PathDelayResponseFollowUp and the PathDelayResponse that we last received. I'm not sure what's going on with the code that deals with the failure of this consistency check, though. According to the state machines in the spec, we should just ignore messages with sequenceIds that don't match our request. If we fail to get a good one before the timeout occurs, the timeout handler should deal with the error condition. For this to work, the PathDelayResponse handler would also have to check the sequenceId and discard if it doesn't match.

To sum up, I don't think we should have to do any of these "sequentiality checks" at all.

You’re right, but… If these sequentiality checks part of the UNH tests then we should either code them or get them removed from the test suite.

The only sequenceId check that matters is the one for equality to the last request message we sent, which we're currently not doing at all!

Checking for equality with the request sequenceId is done here

PTPMessagePathDelayRespFollowUp::processMessage()
…
   if (req->getSequenceId() != sequenceId) {

It should be done in both the response

I explicitly opted not to drop the pdelay response message in the response handler (unless the port isn’t initialized) because it seemed simpler to debug. For example, when a pdelay response followup is received, but there isn’t any response (== NULL) it’s not obvious why.

and follow-up handlers, and non-matches or duplicate messages should count as a lost response. If we don't get a response by the time we are ready to send the next one, then we count that as a lost response too. Whenever we have a lost response, if it's over the threshold we are no longer asCapable; otherwise we proceed to send a request message with the next sequenceId. When we do get a valid response and follow-up, we reset the lost response count. See Figure 11-8 in 802.1AS-2011 for the state machine.

pinealservo:
@christopher-s-hall I assume you replied via email, but unfortunately your email client mangled the quoting. I think if you manually put a > in front of the first line of something you quote, it'll make it apparent that you're quoting the block.

>You’re right, but… If these sequentiality checks part of the UNH tests then we should either code them or get them removed from the test suite.

My point is that *this way* of checking the number of lost packets is problematic; if it's done as I describe, then any reasonable test (i.e. one that won't send responses with sequenceId values we didn't use in previous requests) will pass. The method coded in this branch won't detect missed responses until the next *correct* response arrives, at which point we should be setting asCapable back to true!

> Checking for equality with the request sequenceId is done here
>
> PTPMessagePathDelayRespFollowUp::processMessage()
>…
>   if (req->getSequenceId() != sequenceId) {

That code doesn't exist in the branch under review. It was removed in https://github.com/davidcemin/Open-AVB/commit/01ffc5c70d187211c6c87b35dff153c518b954f1

> I explicitly opted not to drop the pdelay response message in the response handler (unless the port isn’t initialized) because it seemed simpler to debug. For example, when a pdelay response followup is received, but there isn’t any response (== NULL) it’s not obvious why.

I think that's a reasonable decision; I don't think it's critical that we give up on the responses to our pdelay request until it's time to send our next one.

davidcemin:
First, thanks for the discussion. It helps a lot. Some comments below

> To sum up, I don't think we should have to do any of these "sequentiality checks" at all. The only >sequenceId check that matters is the one for equality to the last request message we sent, which we're >currently not doing at all! It should be done in both the response and follow-up handlers, and non->matches or duplicate messages should count as a lost response. If we don't get a response by the time >we are ready to send the next one, then we count that as a lost response too. Whenever we have a lost >response, if it's over the threshold we are no longer asCapable; otherwise we proceed to send a >request message with the next sequenceId. When we do get a valid response and follow-up, we reset >the lost response count.

> My point is that this way of checking the number of lost packets is problematic; if it's done as I describe, >then any reasonable test (i.e. one that won't send responses with sequenceId values we didn't use in >previous requests) will pass. The method coded in this branch won't detect missed responses until the >next correct response arrives, at which point we should be setting asCapable back to true!

What im gathering from this is that, in spite of the fact that the sequential checks work, Levi is proposing another way of checking for errors since this one have some flaws, as for instance not detecting missing responses until the next one arrives. 

> That code doesn't exist in the branch under review. It was removed in davidcemin@01ffc5c
Yes, I did remove that in order to get the different checks from UNH tests to work.

davidcemin:
In order to proceed with this pull request, the suggested path would be accepting Levi's comments and redesign the way we check for lost packets (which will require more invested time on this task) ?

andrew-elder:
@davidcemin - is this pull request now "ready to go" in your opinion?

davidcemin:
@andrew-elder - I've addressed the suggestions raised by @pinealservo and re-tested the whole thing to guarantee that nothing was broken as a consequence. Also, I've removed the unused code that didnt have to be there since we addressed the problem in a different way. I hope I understood correctly what Levi has proposed here. 

The only question still open is the 0.998 and 1.002 topic raised, but I did open a separate issue for that. 
So, answering your question, in my opinion, this pull request is ready to go.

daemons/gptp/common/avbts_message.hpp
```diff
@@ -949,6 +949,8 @@ class PTPMessagePathDelayRespFollowUp:public PTPMessageCommon {
 	PortIdentity *requestingPortIdentity;
 
 	PTPMessagePathDelayRespFollowUp(void) { }
+
+	const static uint16_t SEQID_DIFF_THRESHOLD = 4;
```
pinealservo: This doesn't appear to be used anymore within the diff; is it still needed?  I think there's a similar constant elsewhere that controls the timeout.

daemons/gptp/common/ptp_message.cpp
```diff
@@ -1354,56 +1352,37 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
...
-	{
-		XPTPD_ERROR("Current seqID: %d. %d are missing since the last one sent",
-				sequenceId, sequenceId-port->getLastSeqId());
-		if( (sequenceId - port->getLastSeqId() ) > port->getLostPdelayRespThresh() )
-		{
-			port->setAsCapable( false );
-		}
+	if( req->getSequenceId() != sequenceId ) {
+		XPTPD_ERROR
+			(">>> Received PDelay FUP has different seqID than the PDleay request (%d/%d)",
```
pinealservo: This is just a silly typo, "PDleay" instead of "PDelay", but this would be a great time to fix it.

pinealservo:
I just have a couple of minor and hopefully quick-to-address comments now; otherwise it looks good to me!  Glad to hear that my suggestions worked out.

daemons/gptp/common/avbts_message.hpp
davidcemin: I forgot to remove this one.

daemons/gptp/common/ptp_message.cpp
davidcemin: done. Thanks

pinealservo:
I'm okay with merging it as soon as Travis finishes up.

PR #361, opened 2016-03-20
Halt PDelayReq messages if received multiple responses from different sources

davidcemin:
This pull request adds the PDelayRequest Halt feature when gPTP is receiving multiple responses from different sources. In this case, gPTP is supposed to halt PDelayRequest transmission for 5 minutes. 

Fixes Issue #302

### DIFF
--- a/daemons/gptp/common/ieee1588port.cpp
+++ b/daemons/gptp/common/ieee1588port.cpp
@@ -900,7 +900,7 @@ void IEEE1588Port::processEvent(Event e)
 		break;
 
 	case PDELAY_RESP_PEER_MISBEHAVING_TIMEOUT_EXPIRES:
-		XPTPD_INFO("Timeout expired! Restarting PDelay");
+		XPTPD_INFO("PDelay Resp Peer Misbehaving timeout expired! Restarting PDelay");
 
 		haltPdelay(false);
 		if( port_state != PTP_SLAVE && port_state != PTP_MASTER ) {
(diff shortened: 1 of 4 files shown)


andrew-elder:
This pull request looks good to me. Anyone else?

daemons/gptp/common/ieee1588port.cpp
```diff
@@ -886,6 +898,17 @@ void IEEE1588Port::processEvent(Event e)
 		setAsCapable(false);
 		pdelay_count = 0;
 		break;
+
+	case PDELAY_RESP_PEER_MISBEHAVING_TIMEOUT_EXPIRES:
+		XPTPD_INFO("Timeout expired! Restarting PDelay");
```
pinealservo: Might want to explicitly say which timeout expired, in case someone is looking through logs to investigate "odd" behavior.

pinealservo:
Looks good to me; I left a minor suggestion about a log message. 

Also, 5 minutes is a long time; it seems very reasonable in the absence of any changes, but if someone is watching the output and immediately fixes the problem, it might be nice to have a way to poke it back into action before the timeout expires, or have it reset the timer if the link goes down, etc.

Neither of those need to block the pull request, though; I can file issues for them if you think they're reasonable for future implementation, but I haven't looked at the exact requirements for the behavior being implemented here.

andrew-elder:
Thanks @pinealservo. Agree 5 minutes is a long time, but I think to address it would require reviewing high level requirements documentation. Agree with your approach to file an issue to research requirements.

I'll give @davidcemin a half day to respond to your log message suggestion and will aim to accept the pull request after he makes the minor suggested correction.

davidcemin:
I improved the log message as suggested. Yes, I agree, 5 minutes is a long time. 
We could create a similar procedure (start a timer and count good messages) to restart the PDelay req, or even create a backdoor that would allow the user to send a message to ptp to force it to restore the normal behavior. We just have to make sure that it will continue meeting Avnu requirements.

Issue #254, opened 2015-07-09
gPTP needs to handle link state changes


andrew-elder:
Should handle link up/down and link speed change.

davidcemin:
What is the desirable behaviour for link state changes? For example on linuxptp when the link state changes, all the FSMs are re-started. Should we do something similar for gPTP?

andrew-elder:
Yes, I think so.

davidcemin:
There must be other things to do on the driver level to restart things in case of a fault. With gptp running, i brought the ethernet down (ifconfig ethx down) and I started seeing some error messages. 

After killing gPTP and restarting it, I still get errors like:
Invalid TX
ERROR at 682 in ../../common/ieee1588port.cpp: Error (TX) timestamping PDelay request, error=-72	
ERROR at 1094 in ../../common/ptp_message.cpp: Error (TX) timestamping PDelay Response (Retrying-0), error=-72
ERROR at 1106 in ../../common/ptp_message.cpp: Error (TX) timestamping PDelay Response, error=-72

And these errors only go away if I remove the igb_avb driver and reinsert it again. Im not quite familiar with the igb driver, but is there a way of emptying any FIFOs we have and restarting clean the timestamping side after a fault ? Judging by the error, (-72 is set when reading the tx timestamp returns EAGAIN )

davidcemin:
same happens by changing the linkspeed.

intel-ethernet:
When you ‘down’ then ‘up’ an interface, the gPTP daemon needs to do a series of ioctl calls on the interface to re-enable the desired timestamp settings.

davidcemin:
Could you please list the ioctl calls that need to be done in this case? We would most likely need to call these calls in case of recovering automatically from a failure... (which is the main goal of this issue)

intel-ethernet:
I believe it is basically calling LinuxTimestamperGeneric::HWTimestamper_init() again, but Chris can confirm …

davidcemin:
@christopher-s-hall Can you please confirm latest comment where it says that we could solve this problem by calling  LinuxTimestamperGeneric::HWTimestamper_init() ?

davidcemin:
@kencarlino Did you guys fix this on the automotive branch already ?

kencarlino:
[...]

PR #363, opened 2016-03-28
Handle link up/down events

davidcemin:
Created a link_thread to monitor the NIC status;
Created two extra events on IEEEPort::processEvents to
handle the link up/down state transitions;
On link up, the HWTimestamper init is called. On link down, it is
just being logged the state transition.

This commit is based on the automotive-profile branch and fixes Issue #254.

### DIFF
--- a/daemons/gptp/common/avbts_port.hpp
+++ b/daemons/gptp/common/avbts_port.hpp
@@ -214,6 +214,7 @@ class IEEE1588Port {
 
 	OSNetworkInterface *net_iface;
 	LinkLayerAddress local_addr;
+	int link_delay[4];
 
 	/* Port Status */
 	unsigned sync_count;  // 0 for master, ++ for each sync receive as slave
@@ -287,6 +288,8 @@ class IEEE1588Port {
 	PTPMessageSync *last_sync;
 
 	OSThread *listening_thread;
+
+	OSThread *link_thread;
 
 	OSCondition *port_ready_condition;
 
@@ -473,6 +476,12 @@ class IEEE1588Port {
 	 * @return void
 	 */
 	void recoverPort(void);
+
+	/**
+	 * @brief Watch for link up and down events.
+	 * @return Its an infinite loop. Returns NULL in case of error.
+	 */
+	void *watchNetLink(void);
 
 	/**
 	 * @brief Receives messages from the network interface
@@ -850,6 +859,11 @@ class IEEE1588Port {
 			*msg = '\0';
 		}
 	}
+
+	/**
+	 * @brief Initializes the hwtimestamper
+	 */
+	void timestamper_init(void);
 
 	/**
 	 * @brief  Gets RX timestamp based on port identity
--- a/daemons/gptp/common/ieee1588port.cpp
+++ b/daemons/gptp/common/ieee1588port.cpp
@@ -50,6 +50,17 @@
 LinkLayerAddress IEEE1588Port::other_multicast(OTHER_MULTICAST);
 LinkLayerAddress IEEE1588Port::pdelay_multicast(PDELAY_MULTICAST);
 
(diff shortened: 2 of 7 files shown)


daemons/gptp/common/ieee1588port.cpp
```diff
@@ -145,16 +171,9 @@ bool IEEE1588Port::init_port(int delay[4])
 	this->net_iface = net_iface;
 	this->net_iface->getLinkLayerAddress(&local_addr);
 	clock->setClockIdentity(&local_addr);
+	memcpy(this->link_delay, delay, 4);
```
andrew-elder: What is the intent here? delay is passed in as a parameter of *int delay[4]*?
I assume link_delay is a single int?

daemons/gptp/common/avbts_port.hpp
```diff
@@ -214,6 +214,7 @@ class IEEE1588Port {
 
 	OSNetworkInterface *net_iface;
 	LinkLayerAddress local_addr;
+	int link_delay[4];
```
andrew-elder: 4 ints?

daemons/gptp/common/ieee1588port.cpp
```diff
@@ -134,6 +145,21 @@ IEEE1588Port::IEEE1588Port
 
 	pdelay_count = 0;
 	sync_count = 0;
+	bzero(link_delay, 4);
```
andrew-elder: common code should use memset() instead of bzero()

daemons/gptp/common/ieee1588port.cpp
davidcemin: delay is an array of 4 ints, and it passed as parameter from daemon_cl (both win/lin). The intent here is not to repeat the code. I wanted to call the _hw_timestamper->HWTimestamper_init again when the link would come up, so I wrapped that in another function and created a new private variable that would contain the four integers from the phy delay info.

daemons/gptp/common/avbts_port.hpp
davidcemin: yes, as declared on the function prototype and on daemon_cl.cpp

daemons/gptp/common/ieee1588port.cpp
davidcemin: ok

andrew-elder:
Any more comments on this pull request? I wonder if the link up/down detection conflicts with any changes made on the automotive profile branch? And if there are issues is now the time to address them or not?

davidcemin:
The fixes were manually merged from automotive branch. The implementation itself does not conflict with the one at the automotive branch. What I understood in a quick look in the code is that the link/down events on the automotive branch are protected by an if statement, which is testing if we are running the automotive profile. 

We will definitely have git merge problems between these two branches when we try to merge them together.

daemons/gptp/common/ieee1588port.cpp
kencarlino: Agree that memset should be used. Consider if sizeof(link_delay) should be used rather than 4*sizeof(int).

daemons/gptp/common/ieee1588port.cpp
kencarlino: Can sizeof(delay) be used here on the memcpy()?

kencarlino:
Added a couple of minor comments.

Agreed the merging with the feature-avnu-automotive-profile branch should be a separate task to worry about.

andrew-elder:
@kencarlino - thanks for the comments. I think the "size" comments have been addressed in https://github.com/AVnu/Open-AVB/pull/363/commits/20d08fae4d5b4145971395fd3203ea207a3db111?
Or did you have further comments?

kencarlino:
@andrew-elder I had missed this second commit. This covers the comments I had made. Thanks.

PR #364, opened 2016-03-31
Fix issue with PDelay timer not being set properly

kencarlino:
The Path Delay timer used for issuing Path Delay Request messages was not being set properly.

### DIFF
--- a/daemons/gptp/common/ieee1588port.cpp
+++ b/daemons/gptp/common/ieee1588port.cpp
@@ -230,8 +230,11 @@ bool IEEE1588Port::init_port(int delay[4])
 void IEEE1588Port::startPDelay() {
 	if (automotive_profile) {
 		if (log_min_mean_pdelay_req_interval != PTPMessageSignalling::sigMsgInterval_NoSend) {
+			long long unsigned int waitTime;
+			waitTime = ((long long) (pow((double)2, log_min_mean_pdelay_req_interval) *  1000000000.0));
+			waitTime = waitTime > EVENT_TIMER_GRANULARITY ? waitTime : EVENT_TIMER_GRANULARITY;
 			pdelay_started = true;
-			startPDelayIntervalTimer(log_min_mean_pdelay_req_interval);
+			startPDelayIntervalTimer(waitTime);
 		}
 	}
 	else {


daemons/gptp/common/ieee1588port.cpp
```diff
@@ -230,8 +230,11 @@ bool IEEE1588Port::init_port(int delay[4])
 void IEEE1588Port::startPDelay() {
 	if (automotive_profile) {
 		if (log_min_mean_pdelay_req_interval != PTPMessageSignalling::sigMsgInterval_NoSend) {
+			long long unsigned int waitTime;
+			waitTime = ((long long) (pow((double)2, log_min_mean_pdelay_req_interval) *  1000000000.0));
```
andrew-elder: is log_min_mean_pdelay_req_interval always a positive integer?
pow() using doubles isn't "cheap".
If 1 << X would work, it might be better.

andrew-elder:
Overall, this looks good to me.

daemons/gptp/common/ieee1588port.cpp
kencarlino: log_min_mean_pdelay_req_interval can be and often is a negative value. Also StartPDelay() isn't called to often.

andrew-elder: Fair enough then - thanks for the clarification!

MarcinMiklas: how about using this instead of pow:

waitTime = log_min_pdelay_req_interval<0 ? 1000000000LL>>(-log_min_pdelay_req_interval) : 1000000000LL<<log_min_pdelay_req_interval;

davidcemin:
it looks good to me as well.

daemons/gptp/common/ieee1588port.cpp
kencarlino: The use of pow for the log base 2 calculation for the various timers is used in number places in the gPTP daemon. I'd like to keep them consistent. Perhaps a separate general change would be to pull this out in a separate utility function and update all occurance of this calculation to make use of it.

Issue #243, opened 2015-06-17
correctionField assumes Little-endian Machine. 


davidcemin:
We should be able to detect endianness set this value accordingly.
Code is here:
https://github.com/AVnu/Open-AVB/blob/open-avb-next/daemons/gptp/common/ptp_message.cpp#L522

davidcemin:
I confess that im not sure what I wanted when I opened this issue. I think that the idea is to discuss the way we are storing information within ptp (host/network order). Currently, several internal variables are stored in network order, and changing the whole thing would have a significant cost (code/time). 

Is that what we want to do or can we accept the things the way they are today ?

andrew-elder:
Just fix referenced code, not the whole project.

PR #367, opened 2016-04-01
64 bit words network/host order conversion

davidcemin:
When using 64 bit words, the function byte_swap64 is used. This
function assumes the machine's endianness and because of that it is
not portable.

This patch adds two new functions to deal with the problem of converting
64 bit words between host/network order without assuming the host endianness.

Two byte_swap64 calls were replaced by these two new functions in
PTPMessageCommon::buildCommonHeader and
PTPMessageCommon::buildPTPMessage.

This commit addresses Issue #243


andrew-elder:

christopher-s-hall:
Looks good.

andrew-elder:
Thanks @christopher-s-hall , merging.

PR #370, opened 2016-04-07
Updated the Automotive Profile Readme file

kencarlino:
Added some notes on the implementation that is needed for Windows related to the Automotive Profile changes.


pinealservo:
Since this is just a doc update, I'm going to merge it.

PR #372, opened 2016-04-08
Continue sending announce when priority1 is 255

davidcemin:
If priority1 is set to 255, gPTP must not send any sync
or follow_up messages. gPTP may continue send announce
messages even when priority1 is 255, and BCMA should guarantee
that this device is not a gm.

Reference: IEEE 802.1AS Subclause 10.3.12

### DIFF
--- a/daemons/gptp/common/ieee1588port.cpp
+++ b/daemons/gptp/common/ieee1588port.cpp
@@ -613,7 +613,6 @@ void IEEE1588Port::processEvent(Event e)
 						  (pow((double)2,getSyncInterval())*
 						   1000000000.0)));
 				}
-				return;
 			}
 			if (port_state == PTP_INITIALIZING
 			    || port_state == PTP_UNCALIBRATED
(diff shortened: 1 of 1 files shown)


andrew-elder:
FYI - @davidcemin reran all gPTP tests after this change and they all still passed.

andrew-elder:

christopher-s-hall:
Is the code broken without this change? If not, I'd say don't do it. My memory is foggy, but I think this problem was fixed a few years ago and I'm concerned that this may break something else.

andrew-elder:
@christopher-s-hall - could it be that a few years ago a peer did not correctly implement BCMA? I consider the code "broken" because it does not currently comply with 802.1AS. That said, we certainly don't want to break anything else as a side effect. Having re-read the patch, I can't see how there could be a possible negative side effect. Can you?

andrew-elder:
Per Ken.....

Andrew,
Yes, the automotive profile updates to gPTP account for disabling announce messages independent of anything else that might have been implemented previously in gPTP for disabling announce messages.

andrew-elder:
I'm going to merge because agreement was reached in an email thread external to github.

PR #374, opened 2016-04-16
Need to call igb_detach(...) to accommodate multiple I210's

jpolin:
In pci_connect, which resides in both [examples/common/avb.c](https://github.com/AVnu/Open-AVB/blob/master/examples/common/avb.c) and [examples/simple_talker/simple_talker.c](https://github.com/AVnu/Open-AVB/blob/master/examples/simple_talker/simple_talker.c), there is an edge case which is mis-handled.

The code calls `igb_attach`, then `igb_attach_tx` on each candidate device. If there is more than one I210 card, then igb_attach will succeed on all of them. However, the `igb_attach_tx` only passes if the device is connected. The fail-check does not specifically address the case where the `attach` passes and `attach_tx` fails. Here, we must call `igb_detach` so that future cards will successfully attach. Specifically, we need to reset `pdev->private_data` (by calling `igb_detach`) or the check in `igb_attach` will automatically fail for subsequent calls.


examples/common/avb.c
```diff
@@ -80,6 +80,8 @@ int pci_connect(device_t * igb_dev)
 		err = igb_attach_tx(igb_dev);
 		if (err) {
 			printf("attach_tx failed! (%s)\n", strerror(errno));
+      // If attach passed but attach_tx failed, need to detach
+      if (!err) igb_detach(igb_dev);
```
davidcemin: You should indent the code according to what is already there (tabs/spaces ..)

eric-mann:
Unless I'm mis-reading the merge. ... I don't understand why in a conditional code block ("if (err)") you have code with the negated conditional ("if (!err)"). The code won't execute (?).

andrew-elder:
Joe, thanks for this pull request. Can I ask that you review the comments here and re-submit the request to the "open-avb-next" branch. Our development pattern is that all code changes appear first in the open-avb-next branch before being merged to "master".

I've just realized we don't have the above sequence stated in the readme file. I'll create an issue to have the readme file updated....

examples/common/avb.c
jpolin: Good call, pushed from xcode where it looked fine--will be sure to keep an eye on that in the future!

jpolin:
@eric-mann You're absolutely right, fixed it in simple_talker.c and mindlessly copy/pasted it into the avb.c code. Didn't take the time to notice that the conditional is different.

@andrew-elder Will do, thanks for explaining the process.

DMats:
This error also exists in [avtp_pipeline/openavb_common/avb.c](https://github.com/AVnu/Open-AVB/blob/master/lib/avtp_pipeline/openavb_common/avb.c)

pinealservo:
I'm closing this since it needs some cleanup and to go to open-avb-next instead of master.  I opened issue #396 to track it.

PR #378, opened 2016-04-22
avdecc-lib: add AudioScience avdecc-lib submodule (includes submodule…

andrew-elder:
… jdksavdecc-c as well)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "thirdparty/cpputest"]
 	path = thirdparty/cpputest
 	url = git@github.com:cpputest/cpputest.git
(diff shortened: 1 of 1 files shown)


davidcemin:
looks good to me. So now there will be an effort to implement the full controller using the Audioscience's libs. Do we have the requirements/specs of what we want to be implemented on the controller side for open-avb?

andrew-elder:
I think requirements and specs are still open to discussion. The "who wants to do it" part is also open to discussion.

andrew-elder:
Closing this as we are going to fork avdecc-lib and link our own submodules to the AVnu version of same.

PR #381, opened 2016-04-28
open-avb-next

andrew-elder:
Merge open-avb-next to master


pinealservo:
I've looked through all these.  We apparently let in some formatting inconsistencies, but I think it's probably better to let this merge as-is and fix up formatting again in open-avb-next.

PR #393, opened 2016-05-07
Phase error violation detection added

kencarlino:
Merged changes supplied by Intel to detect phase error violations and restart pdelay state in the case where to many violations occur.

### DIFF
--- a/daemons/gptp/common/ieee1588port.cpp
+++ b/daemons/gptp/common/ieee1588port.cpp
@@ -177,6 +177,8 @@ IEEE1588Port::IEEE1588Port(IEEE1588PortInit_t *portInit)
 	pdelay_count = 0;
 	sync_count = 0;
 
+	_peer_offset_init = false;
+
 	if (testMode) {
 		if (isGM) {
 			avbSyncState = 1;
--- a/daemons/gptp/common/ieee1588port.cpp
+++ b/daemons/gptp/common/ieee1588port.cpp
@@ -59,6 +59,7 @@ OSThreadExitCode watchNetLinkWrapper(void *arg)
 	IEEE1588Port *port;
 
 	port = (IEEE1588Port *) arg;
+	port->link_thread->setName("gPTPLinkWatch");
 	if (port->watchNetlink() == NULL)
 		return osthread_ok;
 	else
@@ -70,8 +71,9 @@ OSThreadExitCode openPortWrapper(void *arg)
 	IEEE1588Port *port;
 
 	port = (IEEE1588Port *) arg;
+	port->listening_thread->setName("gPTPListener");
 	if (port->openPort(port) == NULL)
-		return osthread_ok;
+			return osthread_ok;
 	else
 		return osthread_error;
 }
--- a/daemons/gptp/common/ieee1588port.cpp
+++ b/daemons/gptp/common/ieee1588port.cpp
@@ -73,7 +73,7 @@ OSThreadExitCode openPortWrapper(void *arg)
 	port = (IEEE1588Port *) arg;
 	port->listening_thread->setName("gPTPListener");
 	if (port->openPort(port) == NULL)
-			return osthread_ok;
+		return osthread_ok;
 	else
 		return osthread_error;
 }
--- a/daemons/gptp/common/ptp_message.cpp
+++ b/daemons/gptp/common/ptp_message.cpp
@@ -960,6 +960,13 @@ void PTPMessageFollowUp::processMessage(IEEE1588Port * port)
 		goto done;
 	}
 
+	if (sync->getTimestamp()._version != port->getTimestampVersion())
+	{
+		GPTP_LOG_ERROR
+		    ("Received Follow Up but Sync is out of date");
+		goto done;
+	}
+
 	sync_arrival = sync->getTimestamp();
 
 	delay = port->getLinkDelay();
@@ -1342,6 +1349,7 @@ PTPMessagePathDelayRespFollowUp::~PTPMessagePathDelayRespFollowUp()
 	delete requestingPortIdentity;
 }
 
+#define US_PER_SEC 1000000
 void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
 {
 	Timestamp remote_resp_tx_timestamp(0, 0, 0);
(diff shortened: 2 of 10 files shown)


andrew-elder:
Are these changes automotive specific, or do the fixes just happen to be applied to the automotive branch?

kencarlino:
@andrew-elder They are not specific to the Automotive Profile.

daemons/gptp/common/ptp_message.cpp
```diff
@@ -960,6 +960,13 @@ void PTPMessageFollowUp::processMessage(IEEE1588Port * port)
 		goto done;
 	}
 
+	if (sync->getTimestamp()._version != port->getTimestampVersion())
+	{
+		GPTP_LOG_ERROR
```
pinealservo: We probably should check that the versions match, but this seems like an odd error message to use for mismatched version fields.

daemons/gptp/linux/src/linux_hal_common.hpp
```diff
@@ -180,16 +180,16 @@ class LinuxNetworkInterface : public OSNetworkInterface {
 	( LinkLayerAddress *addr, uint8_t *payload, size_t &length, struct phy_delay *delay );
 
 	/**
-	 * @brief  Disables rx socket descriptor and and clears the rx queue
-	 * @return void
+	 * @brief  Disables rx socket descriptor@return void
```
pinealservo: The `@return` needs to be on the next line.

daemons/gptp/common/ptp_message.cpp
```diff
@@ -1502,12 +1510,23 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
 	    Timestamp prev_peer_ts_theirs;
 	    FrequencyRatio rate_offset;
 	    if( port->getPeerOffset( prev_peer_ts_mine, prev_peer_ts_theirs )) {
+		    FrequencyRatio upper_limit, lower_limit;
+		    upper_limit = UPPER_FREQ_LIMIT*2;
+		    upper_limit /= US_PER_SEC;
+		    upper_limit += 1;
+		    lower_limit = UPPER_FREQ_LIMIT*2;
```
pinealservo: Shouldn't the lower_limit be based on LOWER_FREQ_LIMIT? I am also a little unclear on why we're multiplying by 2 and then adding 1 to each of these.  Would be nice to have a comment explaining what the math is for and where it came from.

daemons/gptp/common/ptp_message.cpp
```diff
@@ -1502,12 +1510,23 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
...
+		    lower_limit /= US_PER_SEC;
+		    lower_limit += 1;
+
 			mine_elapsed =  TIMESTAMP_TO_NS(request_tx_timestamp)-TIMESTAMP_TO_NS(prev_peer_ts_mine);
 			theirs_elapsed = TIMESTAMP_TO_NS(remote_req_rx_timestamp)-TIMESTAMP_TO_NS(prev_peer_ts_theirs);
 			theirs_elapsed -= port->getLinkDelay();
 			theirs_elapsed += link_delay < 0 ? 0 : link_delay;
 			rate_offset =  ((FrequencyRatio) mine_elapsed)/theirs_elapsed;
-			port->setPeerRateOffset(rate_offset);
+
```
pinealservo: Given the above calculation of upper_limit and lower_limit, it seems we're never going to set the rate offset, since upper_limit and lower_limit are always the same value.

daemons/gptp/common/ptp_message.cpp
kencarlino: Enhanced the error message for clarity.

daemons/gptp/linux/src/linux_hal_common.hpp
kencarlino: Will be fixed.

daemons/gptp/common/ptp_message.cpp
kencarlino: Agreed regarding LOWER_FREQ_LIMIT. Good catch. I will forward this for confirmation to the source of these patches at Intel.

daemons/gptp/common/ptp_message.cpp
kencarlino: Depends on result of LOWER_FREQ_LIMIT usage in previous comment.

daemons/gptp/common/ptp_message.cpp
christopher-s-hall: Agreed that the lower_limit should be based on the LOWER_FREQ_LIMIT. The reason to add 1 is:

UPPER_LIMIT = 250 // units of PPM
upper_limit = UPPER_LIMIT/1e6 + 1 
upper_limit = 1.000250 // This is the upper limit of the neighbor rate ratio

The lower limit should be computed:
LOWER_LIMIT = -250 // Again, in PPM units
lower_limit = LOWER_LIMIT/1e6 + 1 
lower_limit = .999750 // This is the lower limit of the neighbor rate ratio.

I originally added the UPPER and LOWER limits to prevent excessive disturbances to the PI controller not the neighbor rate ratio. It seems like a fine idea to limit the neighbor rate ratio, but in reference to the comment above we should probably justify this. The code looks like maybe you want the limit to be +/- 500 PPM?

andrew-elder: What is the advantage of all the complicated math?
Couldn't we just go
upper_limit_ppm = 1.000250  // Upper limit of the neighbor rate ratio in PPM
lower_limit_ppm = 0.999750  // Lower limit of the neighbor rate ratio in PPM

andrew-elder: Oops, it's not really in PPM is it? It is a ratio.... Forget the units part in the above comment.

christopher-s-hall: I think keeping the global/#define in terms of PPM units makes them more intuitive. The math in this particular code maybe less so.

We could add a macro (inline function works too):

#define UPPER_LIMIT_PPM   250
#define LOWER_LIMIT_PPM -250
#define PPM_OFFSET_TO_RATIO(ppm) ((ppm)/((FrequencyRatio)US_PER_SEC)+1)

upper_ratio_limit = PPM_OFFSET_TO_RATIO(UPPER_LIMIT_PPM)
lower_ratio_limit = PPM_OFFSET_TO_RATIO(LOWER_LIMIT_PPM)

pinealservo: I like that last one; PPM-to-ratio conversion is nice and the macro/function makes it clear that's what's happening.

kencarlino:
This pull request is ready to be rechecked.

daemons/gptp/common/ieee1588port.cpp
```diff
@@ -70,8 +71,9 @@ OSThreadExitCode openPortWrapper(void *arg)
 	IEEE1588Port *port;
 
 	port = (IEEE1588Port *) arg;
+	port->listening_thread->setName("gPTPListener");
 	if (port->openPort(port) == NULL)
-		return osthread_ok;
+			return osthread_ok;
```
andrew-elder: Accidental whitespace?

andrew-elder:
Apart from small whitespace glitch, all these changes look good to me.

davidcemin:
Looks good to me.

PR #394, opened 2016-05-07
Fix TLV parse offset

christopher-s-hall:
Previously followup TLV offset was incorrect in the code - the PTP header
length was missing

This patch adds in the length of the PTP header


andrew-elder:
This looks good to me. If anyone else agrees, they can go ahead an merge it.

pinealservo:
Looks good to me too. Merging.

PR #406, opened 2016-05-18
Add daemon start/stop scripts

RikusW:
These are for convenience, I wrote them at work and it might be useful to others as well.

### DIFF
--- a/run_daemons.sh
+++ b/run_daemons.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+# Start all daemons
+
+nic=$1
+ 
+if [ "$1" == "" ];
+then
+    nic="eth1" #edit as required
+    echo "Assuming: "$0" "$nic
+fi
+echo "Starting daemons on "$nic
+
+#use false for silence
+if true; then
+    sudo rmmod igb
+    sudo insmod kmod/igb/igb_avb.ko
+else
+    sudo rmmod igb > /dev/null 2>&1
+    sudo insmod kmod/igb/igb_avb.ko > /dev/null 2>&1
+fi
+
+groupadd ptp
+sudo daemons/gptp/linux/build/obj/daemon_cl $nic &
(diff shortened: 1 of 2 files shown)


andrew-elder:
This looks to be ok to me. Anyone else?

run_daemons.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+# Start all daemons
+
+nic=$1
```
davidcemin: since it is accepting parameters, it would be good to have a -h option that would throw the usage.

davidcemin:
Looks good to me

run_daemons.sh
RikusW: Updated for -h

run_daemons.sh
```diff
@@ -0,0 +1,33 @@
...
+
+if [ "$1" == "-h" ]; then
+        echo "Usage: $0 <network interface>"
+        echo "   eg: $0 eth1"
+        exit
+fi
+
+if [ "$1" == "" ]; then
+        nic="eth1" #edit as required
```
pinealservo: Note for future improvement: It is helpful to have script parameters like this in variables that are defined at the start of the file, which makes it obvious what parameters can be edited and puts them all together.

stop_daemons.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# Stop all daemons
+
+sudo killall maap_daemon
```
pinealservo: For possible future improvement:  Although killall is very convenient, it's not always provided in minimal OS configurations. I don't think there's a clear "best way" to do this across all bash-supporting systems, but there are probably a couple of ways that could be found in some cross-platform start/stop script implementations around.

pinealservo:
I left a couple of comments for possible future improvement, but as they're convenience scripts that aren't required, I'm going to go ahead and merge since @andrew-elder and @davidcemin have already given their approval.

Issue #396, opened 2016-05-12
Need to call igb_detach when igb_attach_tx fails


pinealservo:
See #374, which was submitted to master rather than open-avb-next. The author hasn't opened a pull request against open-avb-next, so someone else needs to take it on, probably along with fixing #395.

RikusW:
Should be fixed since #408

PR #408, opened 2016-05-19
Fix for #396

RikusW:
Fix for #396
pci_connect seems to be duplicated 5 times...

### DIFF
--- a/daemons/gptp/linux/src/linux_hal_i210.cpp
+++ b/daemons/gptp/linux/src/linux_hal_i210.cpp
@@ -81,7 +81,7 @@ pci_connect( device_t *igb_dev )
 		printf("attaching to %s\n", devpath);
 		err = igb_attach(devpath, igb_dev);
 		if (err) {
-			printf("attach failed! (%s)\n", strerror(errno));
+			printf("attach failed! (%s)\n", strerror(err));
 			continue;
 		}
 		/*igb_attach_tx missing here ???*/
--- a/daemons/gptp/linux/src/linux_hal_i210.cpp
+++ b/daemons/gptp/linux/src/linux_hal_i210.cpp
@@ -54,50 +54,41 @@ extern "C" {
 static int
 pci_connect( device_t *igb_dev )
 {
-	struct  pci_access      *pacc;
-	struct  pci_dev *dev;
-	int             err;
-	char    devpath[IGB_BIND_NAMESZ];
+	char devpath[IGB_BIND_NAMESZ];
+	struct pci_access *pacc;
+	struct pci_dev *dev;
+	int err;
 
-	memset( igb_dev, 0, sizeof(device_t));
-
-	pacc    =       pci_alloc();
+	memset(igb_dev, 0, sizeof(device_t));
+	pacc = pci_alloc();
 	pci_init(pacc);
 	pci_scan_bus(pacc);
-	for     (dev=pacc->devices;     dev;    dev=dev->next)
-    {
+
+	for (dev = pacc->devices; dev; dev = dev->next) {
 		pci_fill_info(dev, PCI_FILL_IDENT | PCI_FILL_BASES | PCI_FILL_CLASS);
-      
 		igb_dev->pci_vendor_id = dev->vendor_id;
 		igb_dev->pci_device_id = dev->device_id;
 		igb_dev->domain = dev->domain;
 		igb_dev->bus = dev->bus;
 		igb_dev->dev = dev->dev;
 		igb_dev->func = dev->func;
-		
-		snprintf
-			(devpath, IGB_BIND_NAMESZ, "%04x:%02x:%02x.%d", dev->domain,
-			 dev->bus, dev->dev, dev->func );
-
-		err = igb_probe( igb_dev );
-      
+		snprintf(devpath, IGB_BIND_NAMESZ, "%04x:%02x:%02x.%d",
+				dev->domain, dev->bus, dev->dev, dev->func);
+		err = igb_probe(igb_dev);
 		if (err) {
 			continue;
 		}
-
-		printf ("attaching to %s\n", devpath);
-		err = igb_attach( devpath, igb_dev );
-      
+		printf("attaching to %s\n", devpath);
+		err = igb_attach(devpath, igb_dev);
 		if (err) {
-			printf ("attach failed! (%s)\n", strerror(errno));
+			printf("attach failed! (%s)\n", strerror(errno));
 			continue;
 		}
+		/*igb_attach_tx missing here ???*/
 		goto out;
-    }
-
+	}
 	pci_cleanup(pacc);
 	return  ENXIO;
-  
 out:
 	pci_cleanup(pacc);
 	return  0;
--- a/examples/common/avb.c
+++ b/examples/common/avb.c
@@ -40,7 +40,7 @@
 
 #include "avb.h"
 
-int pci_connect(device_t * igb_dev)
+int pci_connect(device_t *igb_dev)
 {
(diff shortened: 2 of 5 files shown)


RikusW:
In daemons/gptp/linux/src/linux_hal_i210.cpp pci_connect does not call igb_attach_tx, should it ?
The others appear identical, please advise on a common location.

andrew-elder:
We have a large "automotive" pull request pending that I would like to merge before addressing this one.

daemons/gptp/linux/src/linux_hal_i210.cpp
```diff
@@ -54,50 +54,41 @@ extern "C" {
...
-		err = igb_attach( devpath, igb_dev );
-      
+		printf("attaching to %s\n", devpath);
+		err = igb_attach(devpath, igb_dev);
 		if (err) {
-			printf ("attach failed! (%s)\n", strerror(errno));
+			printf("attach failed! (%s)\n", strerror(errno));
 			continue;
 		}
+		/*igb_attach_tx missing here ???*/
```
pinealservo: I'm not positive, but I think agb_attach_tx is only needed when dealing with the userspace interface to transmit on the AVB high-priority queues. I don't think the gPTP implementation uses that interface.

lib/avtp_pipeline/openavb_common/avb.c
```diff
@@ -67,30 +63,27 @@ int pci_connect(device_t * igb_dev)
...
-		if (err)
+		if (err) {
 			continue;
-
+		}
 		printf("attaching to %s\n", devpath);
 		err = igb_attach(devpath, igb_dev);
 		if (err) {
-			printf("attach failed! (%s)\n", strerror(err));
+			printf("attach failed! (%s)\n", strerror(err)); /*errno or err ??*/
```
pinealservo: Looking at the implementation, I would say that `err` is the desired variable.  Should probably be updated in the other files that use `errno`.

It appears to be setting the return value to the value of `errno` already in the case where the APIs called use it to return their errors, and using standard error codes in other cases.

Do @intel-ethernet or @christopher-s-hall have any comments on this?

intel-ethernet: I would only use errno if you are directly calling a system call, not a user-mode library, although the igb library restricted itself to mapping errors to standard system error codes.

andrew-elder: So, the /*errno or err ?? */ comment should be changed to /* use err return from user-mode library, not errno */
Correct?

daemons/gptp/linux/src/linux_hal_i210.cpp
```diff
@@ -54,50 +54,41 @@ extern "C" {
...
 		}
-
-		printf ("attaching to %s\n", devpath);
-		err = igb_attach( devpath, igb_dev );
-      
+		printf("attaching to %s\n", devpath);
+		err = igb_attach(devpath, igb_dev);
 		if (err) {
-			printf ("attach failed! (%s)\n", strerror(errno));
+			printf("attach failed! (%s)\n", strerror(errno));
```
pinealservo: The `errno` here should be `err` instead as per later comment.

examples/common/avb.c
```diff
@@ -67,30 +63,27 @@ int pci_connect(device_t * igb_dev)
...
 		err = igb_probe(igb_dev);
-		if (err)
+		if (err) {
 			continue;
-
+		}
 		printf("attaching to %s\n", devpath);
 		err = igb_attach(devpath, igb_dev);
 		if (err) {
 			printf("attach failed! (%s)\n", strerror(errno));
```
pinealservo: The `errno` instances should be `err` here as well.

examples/jackd-talker/jackd_talker.c
```diff
@@ -213,47 +213,50 @@ void sigint_handler(int signum)
...
+				dev->domain, dev->bus, dev->dev, dev->func);
+		err = igb_probe(igb_dev);
 		if (err) {
 			continue;
 		}
 		printf("attaching to %s\n", devpath);
-		err = igb_attach(devpath, &glob_igb_dev);
+		err = igb_attach(devpath, igb_dev);
 		if (err) {
 			printf("attach failed! (%s)\n", strerror(errno));
```
pinealservo: More `errno` to `err` required here.

examples/jackd-talker/jackd_talker.c
```diff
@@ -446,7 +449,7 @@ int main(int argc, char *argv[])
 		printf("socket creation failed\n");
 		return errno;
 	}
-	err = pci_connect();
+	err = pci_connect(&glob_igb_dev);
 	if (err) {
 		printf("connect failed (%s) - are you running as root?\n",
 		       strerror(errno));
```
pinealservo: This `errno` should be `err` as well.

examples/simple_talker/simple_talker.c
```diff
@@ -358,40 +358,48 @@ void sigint_handler(int signum)
...
+				dev->domain, dev->bus, dev->dev, dev->func);
 		err = igb_probe(igb_dev);
 		if (err) {
 			continue;
 		}
 		printf("attaching to %s\n", devpath);
 		err = igb_attach(devpath, igb_dev);
-		if ( err || igb_attach_tx( igb_dev )) {
+		if (err) {
 			printf("attach failed! (%s)\n", strerror(errno));
```
pinealservo: More `errno` to `err` changes.

lib/avtp_pipeline/openavb_common/avb.c
pinealservo: The comment should probably be removed, and I added comments mentioning where errno should be changed to err.

RikusW:
Just fixed errno to err.

lib/avtp_pipeline/openavb_common/avb.c
```diff
@@ -69,12 +69,12 @@ int pci_connect(device_t *igb_dev)
...
 		err = igb_attach(devpath, igb_dev);
 		if (err) {
-			printf("attach failed! (%s)\n", strerror(err)); /*errno or err ??*/
+			printf("attach failed! (%s)\n", strerror(err));
 			continue;
 		}
 		err = igb_attach_tx(igb_dev);
 		if (err) {
-			printf("attach_tx failed! (%s)\n", strerror(err)); /*errno or err ??*/
+			printf("attach_tx failed! (%s)\n", strerror(err));
```
andrew-elder: I like these changes, but I'm left with the question about what it means to call strerror() with an error returned from a user library. Seems to me there should be a igb_strerror() function since the igb library is the entity that understands what it's returned error codes actually mean.

andrew-elder:
My vote is to accept this pull request as-is and create an issue to review what calling strerror(err) on igb error returns should really be.
Comments?

intel-ethernet:
i agree. 
Merge and start a new thread.

lib/avtp_pipeline/openavb_common/avb.c
christopher-s-hall: The igb library returns system error codes. I don't see anything wrong with this approach. If we want to create our own set of user codes that's fine, but, in my opinion unnecessary. I think an intermediate approach might be to just wrap strerror in igb_strerror() and in the future it could be changed a little more easily.

pinealservo:
I don't see anything wrong with strerror being called on an error value returned by a user library if the library has been designed to use standard error numbers.  It is a bit limiting in the sort of errors you can return, but for things that mostly fail in standard ways, it's nice to not have to redefine a whole alternate set of error codes.

Anyway, looks like consensus is to merge.

Issue #414, opened 2016-05-25
Out-of-bounds read in ClockIdentity::set()


m46kruk:
Hello everyone,

due to improper message-length validation, it is possible for an attacker to cause the gptp daemon to die with a segmentation fault. This happens because the length of an announce message's optional Path Trace TLV is not properly validated in `buildPTPMessage()`.
On top of that, `PathTraceTLV::parseClockIdentity()` relies on the unchecked value of the length field and therefore blindly passes a pointer (that may or may not point to a valid identity) to `ClockIdentity::set()`.

I have attached the code for an application that sends a malformed message with an invalid length (0xFFFF) in the Path Trace TLV's length, which reliably caused daemon_cl to die with a segmentation fault on my side.

I am not sure what the proper process for this is, but I have committed a patch to my fork of this repository. Should I also create a pull request?

[pt_vuln_exploit.zip](https://github.com/AVnu/Open-AVB/files/281947/pt_vuln_exploit.zip)

andrew-elder:
Assuming you have based your fix on the current code in AVnu branch open-avb-next, please do create a pull request to AVnu open-avb-next so that folks can review it. Thanks for finding this and double thanks for fixing it!

m46kruk:
My patch was based on master, but I've fixed it - it's now based on open-avb-new. 
I'm creating the pull request now.

PR #415, opened 2016-05-25
Fix out-of-bounds read in ClockIdentity::set() caused by malformed an…

m46kruk:
Hello,

as discussed in [Issue 414](https://github.com/AVnu/Open-AVB/issues/414), here is my pull request.

Thank you and best regards
Matt

### DIFF
--- a/daemons/gptp/common/avbts_message.hpp
+++ b/daemons/gptp/common/avbts_message.hpp
@@ -360,8 +360,13 @@ class PathTraceTLV {
 	 * @param  buffer [in] Message buffer. It should be at least ::PTP_CLOCK_IDENTITY_LENGTH bytes long.
 	 * @return void
 	 */
-	void parseClockIdentity(uint8_t *buffer) {
-		int length = PLAT_ntohs(*((uint16_t *)buffer))/PTP_CLOCK_IDENTITY_LENGTH;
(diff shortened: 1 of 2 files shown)


daemons/gptp/common/avbts_message.hpp
```diff
@@ -360,8 +360,13 @@ class PathTraceTLV {
 	 * @param  buffer [in] Message buffer. It should be at least ::PTP_CLOCK_IDENTITY_LENGTH bytes long.
 	 * @return void
 	 */
-	void parseClockIdentity(uint8_t *buffer) {
-		int length = PLAT_ntohs(*((uint16_t *)buffer))/PTP_CLOCK_IDENTITY_LENGTH;
+	void parseClockIdentity(uint8_t *buffer, int size) {
```
andrew-elder: the added "size" parameter needs to be added to the doxygen comment block

m46kruk:
Ah sorry, I've added the missing description to the doxygen comment block.
Also, I noticed that I wasn't keeping track of the buffer size correctly, so I've made sure to decrement the size whenever the pointer to the buffer is incremented.

andrew-elder:

davidcemin:
looks good to me

PR #416, opened 2016-05-27
Remove some pci_connect duplicates

RikusW:
Three gptp functions seem to be duplicated as well, with inconsistent return values. I don't have time now, will look into that later.


RikusW:
I had to move the gptp functions to common/ to get the #include "avb.h" to work.
According to grep only simple_talker and jackd_talker used those so the change shouldn't affect anything else.

examples/common/avb.c
```diff
@@ -87,47 +89,59 @@ int pci_connect(device_t *igb_dev)
 	return 0;
 }
 
-int gptpinit(int *shm_fd, char **memory_offset_buffer)
+int gptpinit(int *igb_shm_fd, char **igb_mmap)
 {
-	*shm_fd = shm_open(SHM_NAME, O_RDWR, 0);
-	if (*shm_fd == -1) {
+	if (NULL == igb_shm_fd)
```
RikusW: Checking *igb_mmap here is a good idea.

examples/common/avb.c
```diff
@@ -87,47 +89,59 @@ int pci_connect(device_t *igb_dev)
...
+
+	return 0;
 }
 
-void gptpdeinit(int shm_fd, char *memory_offset_buffer)
+int gptpdeinit(int *igb_shm_fd, char **igb_mmap)
 {
-	if (memory_offset_buffer != NULL) {
-		munmap(memory_offset_buffer, SHM_SIZE);
+	if (NULL != *igb_mmap) {
```
RikusW: Check igb_mmap too ?

RikusW:
I'm happy with it as it is now.

examples/common/avb.h
```diff
@@ -112,11 +112,11 @@ typedef enum { false = 0, true = 1 } bool;
 
 int pci_connect(device_t * igb_dev);
 
-int gptpscaling(gPtpTimeData * td, char *memory_offset_buffer);
+int gptpscaling(char *igb_mmap, gPtpTimeData *td);
```
DMats: Just pointing out that the `gptpscaling()` arguments were reversed to match `jackd_talker` and `simple_talker`.  Did anything else use `gptpscaling()` from `common/avb` ?

RikusW: I used grep, and it seems like only  jackd_talker and simple_talker used it, and it does compile without problems.

DMats:
These changes look good to me.

examples/common/avb.c
```diff
@@ -87,47 +87,72 @@ int pci_connect(device_t *igb_dev)
...
 		shm_unlink(SHM_NAME);
-		return false;
+		return -1;
 	}
-	return true;
+	return 0;
 }
 
-void gptpdeinit(int shm_fd, char *memory_offset_buffer)
+/* gptpdeinit
```
andrew-elder: I like the additional comment. Can we make it Doxygen compatible please....

examples/common/avb.c
```diff
@@ -87,47 +87,72 @@ int pci_connect(device_t *igb_dev)
...
+		ret -= 1;
+	}else if (*igb_shm_fd != -1) {
+		if(close(*igb_shm_fd) == -1) {
+			ret -= 1;
+		}
+		*igb_shm_fd = -1;
 	}
-	if (shm_fd != -1) {
-		close(shm_fd);
+	if (igb_mmap == NULL) {
```
andrew-elder: should this be wrapped in if(0==ret) ?

examples/common/avb.c
RikusW: I don't know doxygen well, will look into that on Monday.

andrew-elder: See pretty much any gPTP header file.... https://github.com/AVnu/Open-AVB/blob/master/daemons/gptp/common/avbts_clock.hpp#L178

examples/common/avb.c
RikusW: igb_mmap should be 0 if not allocated, and if the close above fails for whatever unlikely reason igb_mmap won't be unmapped. (shm_map now)

DMats:
Would it make sense to include the `openavb_common` folder from the `avtp_pipeline` in this consolidation?

Link: https://github.com/AVnu/Open-AVB/tree/master/lib/avtp_pipeline/openavb_common

RikusW:
I have seen the avtp_pipeline one, it might be better to finish this and decide later.

RikusW:
I'd like to squash some of these together, but aren't sure whether it is a good idea...

andrew-elder:
@RikusW - please do go ahead and squash what you think is best. Until your branch is merged in here, it is essentially yours to do what you want with. Now is the time to do it!

RikusW:
@andrew-elder squashing is done, the other common to lib/common files can be done in another branch there is already too much in here.

andrew-elder:
These changes look good to me, although I'm not in a position to test them out. Anyone else have comments?

DMats:
Visually, the changes look good to me too.  But, I never tested most of the previous code and won't get the chance to test this new stuff for at least 2 weeks.  I'd say move forward with the changes.  Consolidation will make it easier to shake out any remaining bugs, if they exist.

I'm in favor of consolidating the other common files and functions.  I spent many hours looking at different versions of `mrp_get_domain` that could have been saved had there only been 1 version.

lib/common/Makefile
```diff
@@ -0,0 +1,12 @@
+CC ?= gcc
```
pinealservo: @andrew-elder Is this the file/line you mentioned causing a build problem for you in some context? I don't remember the details of what you said or if it's something that needs fixing before we merge.

pinealservo:
I am all for this consolidation work; it does a lot to tidy up the unnecessary duplication and bring more core functionality together into library form, which I believe will be helpful to many. I am currently doing some work on the avtp_pipeline code, so I can test the changes made there.  I'll comment again after I've tried it out.

lib/common/Makefile
andrew-elder: @pinealservo - yes it does cause issues for us at present, but I think we should go ahead and merge anyway. We can discuss the correct approach separately. We have the option of maintaining our own repo/patches/branches....

lib/common/avb.c
```diff
@@ -87,47 +93,104 @@ int pci_connect(device_t *igb_dev)
...
+ * @brief Free the memory mapping used for IPC
+ * @param shm_fd [in] File descriptor for mapping
+ * @param shm_map [in] Pointer to mapping
+ * @return 0 for success, negative for failure
+ *	-1 = close failed
+ *	-2 = munmap failed
+ *	-3 = close and munmap failed
+ */
+
+int gptpdeinit(int *shm_fd, char **shm_map)
```
pinealservo: This has a different signature than the one that was in avtp_pipeline; there's an extra pointer indirection for each argument. There's a call in openavb_time_osal.c that needs to be updated.  You should see some warnings related to it when you build it.

RikusW: @pinealservo I did update that, look right at the end of my last commit.

RikusW: @pinealservo Fixed.

pinealservo:
I've merged this with the avtp_pipeline code I've been working on, and it required some additional changes to the avtp_pipeline integration to work, but it's really close.  I'll make some additional comments in-line.

lib/avtp_pipeline/platform/Linux/openavb_time_osal.c
```diff
@@ -71,7 +71,7 @@ static bool x_timeInit(void) {
 		return FALSE;
```
pinealservo: gptpinit error reporting sense has changed; the code here needs to be changed to check for < 0 rather than 0.

lib/avtp_pipeline/platform/Linux/openavb_time_osal.c
```diff
@@ -88,7 +88,7 @@ static bool x_timeInit(void) {
 static bool x_getPTPTime(U64 *timeNsec) {
 	AVB_TRACE_ENTRY(AVB_TRACE_TIME);
 
-	if (!gptpscaling(&gPtpTD, gIgbMmap)) {
+	if (!gptpscaling(gIgbMmap, &gPtpTD)) {
```
pinealservo: gptpscaling calls need to also check for < 0 instead of 0.

lib/common/avb.c
```diff
@@ -87,47 +93,104 @@ int pci_connect(device_t *igb_dev)
...
+ * @return 0 for success, negative for failure
+ */
+
+int gptpscaling(char *shm_map, gPtpTimeData *td) //change this function name ??
+{
+	int i;
+	if ((i = gptpgetdata(shm_map, td)) < 0) {
+		return i;
+	}
 	fprintf(stderr, "ml_phoffset = %" PRId64 ", ls_phoffset = %" PRId64 "\n",
```
pinealservo: We need to kill these fprintfs. The pipeline code calls it WAY too often to have debug output like this.

pinealservo:
With the changes I just noted, I think we'll be in good shape to merge.

lib/common/avb.c
RikusW: I created gptpgetdata to avoid the printfs, any reason it is called gptpscaling ?

andrew-elder:
@pinealservo, I see an update from @RikusW. Are we "good to go" now?

pinealservo:
It’s probably good to go; I haven’t tested it again yet, but I can fix any remaining problems with the avtp_pipeline integration if there are any.

PR #420, opened 2016-06-07
Memory leak in buildPTPMessage()

m46kruk:
Hello,

this patch fixes a memory leak in `buildPTPMessage()`:
When a message with unset transport_specific flag is received, `buildPTPMessage()` will immediately return NULL, without freeing any previously allocated resources.
This patch changes the function so that the check is performed before a PortIdentity is allocated, and instead of returning NULL immediately, jumps to the cleanup code at the end of the function (which is also used in case of other error conditions).

To verify the memory leak and the fix, the attached source can be used (the code will send 1,000,000 announce message with unset transport_specific flag).
[unspecific_memory_leak.zip](https://github.com/AVnu/Open-AVB/files/302053/unspecific_memory_leak.zip)

### DIFF
--- a/daemons/gptp/common/ptp_message.cpp
+++ b/daemons/gptp/common/ptp_message.cpp
@@ -98,6 +98,11 @@ PTPMessageCommon *buildPTPMessage
 	messageType = (MessageType) (tspec_msg_t & 0xF);
 	transportSpecific = (tspec_msg_t >> 4) & 0x0F;
 
+	if (transportSpecific!=1) {
+		GPTP_LOG_EXCEPTION("*** Received message with unsupported transportSpecific type=%d",transportSpecific);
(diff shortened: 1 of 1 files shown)


andrew-elder:
This change looks good to me. Anyone else?

christopher-s-hall:
Looks OK to me. I realize that is what was there, but the formatting could be improved a bit. More like:

if (transportSpecific != 1) {

andrew-elder:
Or even, if (1 != transportSpecific) {

m46kruk:
Thanks, I have improved the formatting according to your suggestions

pinealservo:
Looks good to me.  Since the request made by the other reviewers was completed, I'm going to merge it.

PR #423, opened 2016-06-10
Change all doxygen-style comments to have a @brief

pinealservo:
As promised in the automotive profile pull request review, I've put together a patch that fixes all the Doxygen-style comments that didn't have any tags at all. They now all have at least a @brief tag. There were just a couple that got a slight bit of extra rewording, but this is mostly a mechanical change.

There are also two or three places where I fixed the whitespace in close proximity to one of these comment blocks.

I found there was a common theme: C++ constructors and destructors almost universally had Doxygen-style comment blocks with no @brief or other tags. If this was intentional, maybe a better fix for those would be to turn them into regular comments instead of Doxygen comments with @brief? The Doxygen output with this patch in place doesn't look objectionable to me, although I haven't gone through all of it.

### DIFF
--- a/daemons/gptp/common/avbap_message.hpp
+++ b/daemons/gptp/common/avbap_message.hpp
@@ -66,7 +66,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 /**
- * Automotive Profile Test Status Station State
+ * @brief Automotive Profile Test Status Station State
  */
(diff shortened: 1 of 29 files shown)


andrew-elder:
My guess is that lack of @brief in the C++ constructors and destructors was an oversight rather than intentional. I've scanned the changes and all looks good to me. I'm going to go ahead and merge this pull request without waiting for additional input because there is no functional change to test and review.

PR #424, opened 2016-06-20
Fix logic for starting PDelay sending

pinealservo:
The else clause was attached to the wrong if condition; as a result, sending of PDelay messages only worked when the automotive profile was enabled. Without PDelay transactions, the peer would never be asCapable and the rest of the protocol would never run.

I have verified that it does send PDelay requests when this patch is applied.

### DIFF
--- a/daemons/gptp/common/ieee1588port.cpp
+++ b/daemons/gptp/common/ieee1588port.cpp
@@ -241,10 +241,10 @@ void IEEE1588Port::startPDelay() {
 				pdelay_started = true;
 				startPDelayIntervalTimer(waitTime);
 			}
-			else {
-				pdelay_started = true;
-				startPDelayIntervalTimer(32000000);
-			}
+		}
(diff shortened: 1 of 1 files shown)


andrew-elder:
Looks good to me. David?

DMats:
Looks good to me as well.

andrew-elder:
@DMats - someone is going to test the patch a little tomorrow before we merge.

davidcemin:
The code looks good to me too. The announce messages are being received correctly now. I think we can go ahead and integrate this and if I find more problems we fix them as they appear. I'll leave tests running overnight

davidcemin:
If we can put this off till tomorrow it would be great, so I will leave it running overnight.

davidcemin:
There are some errors but I judge that they should be addressed separately.  I will go ahead and integrate it since open-avb-next is broken today without it.

PR #425, opened 2016-06-21
Fix default initialLogPdelayReqInterval

pinealservo:
The automotive profile code merge introduced configuration parameters for the various interval timers, but unfortunately it set the initial Pdelay Request timer default incorrectly.  This patch fixes the problem.


andrew-elder:
This change is so simple and obvious I'm just going to go ahead an merge it.

PR #426, opened 2016-06-22
Memory leak

afx00:
The first patch fixes memory leak that occurs when receive unexpected message in PTP packet.
In addition,  checking transportSpecifit needs to be done after calling getRxTimestamp().
If buildPTPMessage() is returned before calling getRxTimeStamp() rxTimestampList will keep increasing in LinuxTimestamperGeneric class.

The second patch removes to assign duplicated InterfaceName object unnecessarily.
And this also occurs memory leak.


andrew-elder:
These changes look ok to me. Anyone else?

PR #428, opened 2016-06-28
Change gptpscaling calls to gptpgetdata to avoid printfs

pinealservo:
The version of gptpscaling in the new centralized avb library is just a wrapper around gptpgetdata with some printfs; the avtp_pipeline library doesn't need the printfs, so I have changed it to use gptpgetdata instead.

### DIFF
--- a/lib/avtp_pipeline/platform/Linux/openavb_time_osal.c
+++ b/lib/avtp_pipeline/platform/Linux/openavb_time_osal.c
@@ -71,8 +71,8 @@ static bool x_timeInit(void) {
 		return FALSE;
 	}
 
-	if (gptpscaling(gIgbMmap, &gPtpTD) < 0) {
-		AVB_LOG_ERROR("GPTP scaling failed");
+	if (gptpgetdata(gIgbMmap, &gPtpTD) < 0) {
+		AVB_LOG_ERROR("GPTP data fetch failed");
 		AVB_TRACE_EXIT(AVB_TRACE_TIME);
 		return FALSE;
 	}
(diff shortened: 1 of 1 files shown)


RikusW:
@pinealservo You might want to change this too in examples/

./jackd-talker/jackd_talker.c:	if (-1 == gptpscaling(igb_mmap, &td)) {
./simple_talker/simple_talker.c:	if (-1 == gptpscaling(igb_mmap, &td)) {

pinealservo:
@RikusW If I change both of those, then nobody will be calling gptpscaling, and we might as well have had you just remove the printfs from gptpscaling. The jackd_talker.c and simple_talker.c used to have the printfs, didn't they?  The avtp_pipeline calls didn't, so I'm just restoring that state of things.

I don't know if we want printfs in the examples or not, but I know we don't want them and didn't have them previously in avtp_pipeline.

RikusW:
@pinealservo is there any good reason to call it gptpscaling instead of gptpgetdata which seems to me to be more descriptive of what is actually happening ?

andrew-elder:
@RikusW - the patch changes the calls to gptpgetdata(), so I think we are good right? If you wish to change the examples, we should do that in a separate pull request.

I'm ready to merge this one. I need one more person to sign-off on it. It's only 4 lines to review, so please take a quick look and leave a comment.

kencarlino:
Looks good to me.

RikusW:
Looks good to me too, just saw the comment when I got to work this morning, I'm at GMT+2.

andrew-elder:
Thanks for the feedback @RikusW !

PR #431, opened 2016-07-11
Updates to gPTP

kencarlino:
This pull request updates a number of items in gPTP in a series of commits, including.

Fixes to Syntonize mode to work with all sync rates rather than just sync rates of 125ms
Fixes to some Automotive Profile merge issues
Addition of the Phase Error Violation Recovery patches
Additional status logging when the -E option is enabled
Some cleanup to logging to remove extra newlines

Regards,
Ken

### DIFF
--- a/daemons/gptp/common/avbts_clock.hpp
+++ b/daemons/gptp/common/avbts_clock.hpp
@@ -44,6 +44,7 @@
 
 #define EVENT_TIMER_GRANULARITY 5000000		/*!< Event timer granularity*/
 
+/* These 4 macros are used only when Syntonize mode is enabled */
 #define INTEGRAL 0.0003				/*!< PI controller integral factor*/
 #define PROPORTIONAL 1.0			/*!< PI controller proportional factor*/
 #define UPPER_FREQ_LIMIT  250.0		/*!< Upper frequency limit */
--- a/daemons/gptp/common/avbts_clock.hpp
+++ b/daemons/gptp/common/avbts_clock.hpp
@@ -44,7 +44,7 @@
 
 #define EVENT_TIMER_GRANULARITY 5000000		/*!< Event timer granularity*/
 
-#define INTEGRAL 0.0024				/*!< PI controller integral factor*/
+#define INTEGRAL 0.0003				/*!< PI controller integral factor*/
 #define PROPORTIONAL 1.0			/*!< PI controller proportional factor*/
 #define UPPER_FREQ_LIMIT  250.0		/*!< Upper frequency limit */
 #define LOWER_FREQ_LIMIT -250.0		/*!< Lower frequency limit */
--- a/daemons/gptp/common/avbts_clock.hpp
+++ b/daemons/gptp/common/avbts_clock.hpp
@@ -49,6 +49,18 @@
 #define UPPER_FREQ_LIMIT  250.0		/*!< Upper frequency limit */
 #define LOWER_FREQ_LIMIT -250.0		/*!< Lower frequency limit */
 
+#define UPPER_LIMIT_PPM 250
+#define LOWER_LIMIT_PPM -250
+#define PPM_OFFSET_TO_RATIO(ppm) ((ppm) / ((FrequencyRatio)US_PER_SEC) + 1)
+
+/* This is the threshold in ns for which frequency adjustments will be made */
+#define PHASE_ERROR_THRESHOLD (1000000000)
+
+/* This is the maximum count of phase error, outside of the threshold before
+   adjustment is performed */
+#define PHASE_ERROR_MAX_COUNT (6)
+
+
 /**
  * @brief Provides the clock quality abstraction.
  * Represents the quality of the clock
@@ -106,6 +118,7 @@ private:
 	bool _syntonize;
 	bool _new_syntonization_set_point;
 	float _ppm;
+	int _phase_error_violation;
 
 	IEEE1588Port *port_list[MAX_PORTS];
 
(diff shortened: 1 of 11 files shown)


daemons/gptp/common/ieee1588clock.cpp
```diff
@@ -350,26 +358,44 @@ void IEEE1588Clock::setMasterOffset
...
 		// Adjust for frequency offset
 		long double phase_error = (long double) -master_local_offset;
-		_ppm += (float) (INTEGRAL*phase_error +
-			 PROPORTIONAL*((master_local_freq_offset-1.0)*1000000));
+		if( fabsl(phase_error) > PHASE_ERROR_THRESHOLD ) {
+			++_phase_error_violation;
+		} else {
+			_phase_error_violation = 0;
+
+			float syncPerSec = (float)(1.0 / pow((float)2, port->getSyncInterval()));
```
andrew-elder: Maybe we should add a getSyncPerSec() method to the port? Seems like overkill to call a pow() operation every time this runs - although I realize it doesn't run that often. I don't consider this a "blocking" comment - it's just a suggestion.

daemons/gptp/common/ieee1588port.cpp
```diff
@@ -107,7 +107,7 @@ IEEE1588Port::IEEE1588Port(IEEE1588PortInit_t *portInit)
 		if (initialLogSyncInterval == LOG2_INTERVAL_INVALID)
 			initialLogSyncInterval = -5;     // 31.25 ms
 		if (initialLogPdelayReqInterval == LOG2_INTERVAL_INVALID)
-			initialLogPdelayReqInterval = -3;  // 125 ms
```
pinealservo: Was this perhaps meant to be the operLogSyncInterval instead?  Having the sync interval at 1 second actually seems to cause problems too.

daemons/gptp/linux/src/daemon_cl.cpp
```diff
@@ -99,6 +99,8 @@ static IEEE1588Port *pPort = NULL;
 
 int main(int argc, char **argv)
 {
+	GPTP_LOG_INFO("gPTP starting");
```
pinealservo: For some reason, this file is indented by spaces rather than tabs, so this line does not match the formatting of the rest of the file. Probably best to keep file-level consistency and possibly switch this file back to consistency with the rest of the gptp code another time.

daemons/gptp/linux/src/linux_hal_common.hpp
```diff
@@ -180,17 +180,17 @@ class LinuxNetworkInterface : public OSNetworkInterface {
...
 
 	/**
-	 * @brief  Enables the rx socket descriptor
+	 * @brief  Clears and enables the rx socket descriptor
 	 * @return void
 	 */
-	void reenable_rx_queue();
-
+	void clear_reenable_rx_queue();
+
```
pinealservo: Extra whitespace on this line.

daemons/gptp/common/ptp_message.cpp
```diff
@@ -1611,23 +1616,31 @@ void PTPMessagePathDelayRespFollowUp::processMessage(IEEE1588Port * port)
...
-					port->setAsCapable( true );
-				}
+
+			if( rate_offset < upper_ratio_limit && rate_offset > lower_ratio_limit ) {
+				port->setPeerRateOffset(rate_offset);
 			}
 		}
 	}
-
+	if( !port->setLinkDelay( link_delay ) ) {
```
pinealservo: Good catch; I guess this resulted in some incorrect calculations when running in automotive profile mode?

daemons/gptp/common/avbts_clock.hpp
```diff
@@ -44,11 +44,23 @@
 
 #define EVENT_TIMER_GRANULARITY 5000000		/*!< Event timer granularity*/
 
-#define INTEGRAL 0.0024				/*!< PI controller integral factor*/
+#define INTEGRAL 0.0003				/*!< PI controller integral factor*/
 #define PROPORTIONAL 1.0			/*!< PI controller proportional factor*/
 #define UPPER_FREQ_LIMIT  250.0		/*!< Upper frequency limit */
 #define LOWER_FREQ_LIMIT -250.0		/*!< Lower frequency limit */
 
+#define UPPER_LIMIT_PPM 250
```
pinealservo: I guess that these new macros apply just to the syntonization mode, so it might be helpful to have a comment saying as much if that's really the case.

daemons/gptp/common/ieee1588clock.cpp
kencarlino: As far as I know this is the only usage of a sync per second. If it is being done in other places I would agree also that it could be moved to the port class.
There was previously some discussion about the use of pow() to resolve the base 2 intervals. This pattern is used in all occurrences in gPTP. For consistency I would favor to keep them all the same. That said if someone wanted to go and update the usage throughout gPTP to something else I would support that as long as it remains consistent. Along with that perhaps it would make sense if someone does do that to add interfaces to the port class to return either the base 2 value or interval in time (ms, us or ns).

daemons/gptp/common/ieee1588port.cpp
kencarlino: Initial PDelayRequestInterval of 1 second is according to the Automotive Profile. There are two other commits in this pull request the impact the correct functioning of PDelay responses arriving before a first sync as well as the sync intervals of anything different than 125ms when Syntonize mode is in use. For clarity the intent of the defaults are:
Init Sync: 31.25 ms
Oper Sync: 1 second
Init PDelay Req: 1  second
Oper PDelay Req: 1 second

daemons/gptp/linux/src/daemon_cl.cpp
kencarlino: Converted daemon_cl.cpp whitespace to be tabs rather than spaces.

daemons/gptp/common/ptp_message.cpp
kencarlino: I noticed this when I was troubleshooting the clock frequency adjustment issue so I can't say the level of impact of this but the link delay is used in calculations so yes it would lead to some error in calculations.

daemons/gptp/common/avbts_clock.hpp
kencarlino: Added a comment.

pinealservo:
Looks good to me now.

PR #434, opened 2016-07-19
gPTP Fixes and Improvements

kencarlino:
Here is another set of fixes and improvements for gPTP with the focus on the automotive profile. See the list of commits for more information about what is changed.

### DIFF
--- a/daemons/gptp/common/ieee1588port.cpp
+++ b/daemons/gptp/common/ieee1588port.cpp
@@ -187,7 +187,7 @@ IEEE1588Port::IEEE1588Port(IEEE1588PortInit_t *portInit)
 		}
 		linkUpCount = 1;  // TODO : really should check the current linkup status http://stackoverflow.com/questions/15723061/how-to-check-if-interface-is-up
 		linkDownCount = 0;
-		stationState = STATION_STATE_RESERVED;
+		setStationState(STATION_STATE_RESERVED);
 	}
 }
 
@@ -572,7 +572,7 @@ void IEEE1588Port::processEvent(Event e)
 
 		if (automotive_profile) {
 			if (testMode) {
-				stationState = STATION_STATE_ETHERNET_READY;
+				setStationState(STATION_STATE_ETHERNET_READY);
 				APMessageTestStatus *testStatusMsg = new APMessageTestStatus(this);
 				if (testStatusMsg) {
 					testStatusMsg->sendPort(this);
@@ -712,7 +712,7 @@ void IEEE1588Port::processEvent(Event e)
 			asCapable = true;
 
 			if (testMode) {
-				stationState = STATION_STATE_ETHERNET_READY;
+				setStationState(STATION_STATE_ETHERNET_READY);
 				APMessageTestStatus *testStatusMsg = new APMessageTestStatus(this);
 				if (testStatusMsg) {
 					testStatusMsg->sendPort(this);
@@ -973,8 +973,8 @@ void IEEE1588Port::processEvent(Event e)
 					if (avbSyncState > 0) {
 						avbSyncState--;
 						if (avbSyncState == 0) {
-							// Send an initial signalling message
-							stationState = STATION_STATE_AVB_SYNC;
+							// Send Avnu Automotive Profile status message
+							setStationState(STATION_STATE_AVB_SYNC);
 							APMessageTestStatus *testStatusMsg = new APMessageTestStatus(this);
 							if (testStatusMsg) {
 								testStatusMsg->sendPort(this);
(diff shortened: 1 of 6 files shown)


andrew-elder:
This looks ok to me. Is anyone in a position to run it?

pinealservo:
I am (will be) in a position to run it; I'm in the middle of a long-run test right now, so I'll try it tomorrow.

daemons/gptp/common/ieee1588port.cpp
```diff
@@ -1251,10 +1274,12 @@ void IEEE1588Port::becomeSlave( bool restart_syntonization ) {
 
 	port_state = PTP_SLAVE;
 
-	/*clock->addEventTimer
-		( this, SYNC_RECEIPT_TIMEOUT_EXPIRES,
-		(SYNC_RECEIPT_TIMEOUT_MULTIPLIER*
-		 (unsigned long long)(pow((double)2,getSyncInterval())*1000000000.0)));*/
+	/* TODO : Seems this be active here
```
pinealservo: I don't know why we would ever want a sync receipt timer running while we're grandmaster.  Should probably be deleted unless someone knows a good reason for it to continue to stick around as a comment.

daemons/gptp/common/ieee1588port.cpp
```diff
@@ -1230,10 +1251,12 @@ void IEEE1588Port::getDeviceTime
 
 void IEEE1588Port::becomeMaster( bool annc ) {
 	port_state = PTP_MASTER;
-	// Start announce receipt timeout timer
-	// Start sync receipt timeout timer
+	// Stop announce receipt timeout timer
 	clock->deleteEventTimer( this, ANNOUNCE_RECEIPT_TIMEOUT_EXPIRES );
```
pinealservo: This doesn't need attention as part of this pull request, but these changes made me wonder if we ought to update the management of the announce receipt timer to follow the same pattern the others do now, with helper functions and locks.

daemons/gptp/common/ieee1588port.cpp
```diff
@@ -1274,17 +1276,20 @@ void IEEE1588Port::becomeSlave( bool restart_syntonization ) {
 
 	port_state = PTP_SLAVE;
 
-	/* TODO : Seems this be active here
+	/* TODO : Should this be active for non-automotive profile uses
```
pinealservo: These commented out things relating to starting the sync timeout are apparently from [this commit](https://github.com/AVnu/Open-AVB/commit/32d59d82495d2aabda9c952016a6a2dce372be90l), which is supposed to wait for the first sync receive to start.  There is indeed code to start the SYNC_RECEIPT timer when a Sync Follow-Up is received by a clock slave.

I'm not sure this is technically correct, but it would take a strangely malfunctioning master for it to cause a problem and it's probably convenient in some cases where you want to force a device to wait for the master to arrive rather than take over as temporary master.

We should probably delete the commented out code at some point, unless someone objects to this behavior.

pinealservo:
Ugh, my comments aren't really on outdated diffs; I was just viewing a commit at a time.  In any case, I didn't find anything that would prevent merging, I just looked into some of the "TODO" commented out code, so you might want to un-collapse and look at my previous comments.

I have run it in both normal and automotive profile mode, and it seems to be working correctly, though I haven't run exhaustive tests.

daemons/gptp/common/ieee1588port.cpp
kencarlino: I agree with regards to the announce receipt timer and considered updating it but decided better to not change code I wasn't going to be in a position to test.

daemons/gptp/common/ieee1588port.cpp
kencarlino: Also I agree with this comment but decided to just add the TODO so that it could be looked at in more detail when someone has a chance.

christopher-s-hall: Not starting the sync receipt timer is the "right thing". I think there is (or at least was) a compliance test for this and there may be failures if we make a change and start the timer here. This is the correct interpretation of the state machine in the 802.1AS standard, but the state machine may not represent the correct or intended behavior.

kencarlino: Seeing that discussion continued on this I also checked the spec and concur that not starting the sync receipt timer until a sync is received and the slave knows what rate to expect is the right thing to do. I have gone ahead and removed the TODO comment I had added and removed the commented out code as well to reduce confusion in the future.

pinealservo:
Thanks for cleaning up that comment, @kencarlino.  I think we've had enough eyes on this now, and I've verified basic functionality in both standard and automotive profile mode.

andrew-elder:
Thanks for merging @pinealservo!

PR #439, opened 2016-07-25
Removing redundant line causing AVB streamhandler writeback error

jfornal:
This change prevents from receiving kernel warnings ("map pfn RAM range req uncached-minus for[...]"). It is unecessary to call that method within igb_mmap(). The behaviour of the code after that change has been tested and proved to be working (no more kernel warnings and audio playback possible)

### DIFF
--- a/kmod/igb/igb_main.c
+++ b/kmod/igb/igb_main.c
@@ -10397,7 +10397,6 @@ static int igb_mmap(struct file *file, struct vm_area_struct *vma)
 	} else {
 		physaddr = pgoff;
 	}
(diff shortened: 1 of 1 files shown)


andrew-elder:
Thanks @jfornal. Looks ok to me. Any comments from Intel folds that know more about Linux driver interfacing than I do?

jfornal:
Just to let you know: It's been consulted with Eric Mann (actually the change has been proposed by him).

eric-mann:
Change looks good to me !

andrew-elder:
Thanks all, I'm going to merge this.

PR #441, opened 2016-07-26
Open avb next

andrew-elder:
gPTP updates


andrew-elder:
I'm opening this pull request because it is overdue. If there are reasons we shouldn't do it at this time, now is the time to raise any concerns.....

pinealservo:
As far as I'm aware, there aren't any issues this would introduce that aren't already in master, but we might want to fix the issue with calling timestamper_init multiple times first.  If we don't do it first, we should do another merge shortly after that fix lands in open_avb_next.

andrew-elder:
My plan is to leave the comment Window open until Thursday and if we don't receive any more comments, go ahead and complete the merge.

PR #443, opened 2016-07-27
MRP: bugfix, mrpd.c needs to send strings starting with 'I' to msrp

andrew-elder:


daemons/mrpd/mrpd.c
```diff
@@ -331,6 +331,7 @@ int process_ctl_msg(char *buf, int buflen, struct sockaddr_in *client)
 		return mvrp_recv_cmd(buf, buflen, client);
 		break;
 	case 'S':
+	case 'I':
```
pinealservo: Would you mind adding the following lines from `daemons/mrpd/msrp.c` to the comment above this code?

```
* I+S   Add a stream id to the talker stream id list
* I-S   Remove a stream id from the talker stream id list
* I-A   Remove all stream ids from the interesting talker and listener stream id lists
```

I think those comment blocks are currently the primary documentation for the protocol, so it would be good to keep them up-to-date.

Otherwise, looks good!

andrew-elder: Good catch - thanks. Done!

pinealservo:
I think this is straightforward enough to merge; you can go ahead unless you'd like to wait for after the open-avb-next merge. Doing it before would be my preference, though.

andrew-elder:
Done

PR #444, opened 2016-07-27
Bug fixes for gPTP daemon

k-gupta:
This pull request includes two bug fixes to the gPTP daemon. Following is a detailed explanation for the bugs observed and their fixes:

**0d1173c: gPTP: Fix end node adjusting time in PTP_INITIALIZING state**

This issue was observed when two devices (A and B) running the gPTP daemon are directly connected. However, this issue also affects a scenario where multiple devices are connected to a grandmaster on a time-aware 802.1AS network. 
 
When the gPTP daemon is started, both the devices set asCapable as enabled and device A is chosen as the grandmaster by the BMC algorithm. If device A goes offline when device B is adjusting its freq, device B goes into the PTP_MASTER state and sets asCapable as disabled but retains the current freq adjustment. When device A comes back online, both devices set asCapable as enabled and device A goes into PTP_INITIALIZING state. Since device B is currently the GM, device A receives Sync and FollowUp messages from B and starts adjusting its clock to device B's frequency. On the next ANNOUNCE_TIMEOUT event, device A is chosen as the new GM by the BMC algorithm but its clock is running at an incorrect frequency. Since it is a GM, it never updates/resets its clock frequency. Therefore, device B starts driving its clock based on this incorrect frequency. 
 
Depending on the magnitude of error in the grandmaster's frequency, it is possible that the target frequency the slave should achieve in order to synchronize to the grandmaster might lie outside the slave's adjustment range, making it impossible to stay synchronized to the grandmaster.

To fix this issue, the code has been changed so that the end nodes adjust their clocks only in the PTP_SLAVE state. 

**318c3f4: gPTP: Fix correction applied to preciseOriginTimestamp**

This issue was observed when two devices running the gPTP daemon are connected through a switch with the syntonize option enabled (-S). The phase offset on the slave device oscillates in the range of tens of microseconds and never converged. 
During the investigation, it was found that when processing a FollowUp message, the correction to the preciseOriginTimestamp is being calculated incorrectly. The correctionField is converted into the local time base using the rate ratio and added to the path delay to calculate the correction value (in the local time base). This correction is then applied to the preciseOriginTimestamp (which is in the grandmaster time base). This preciseOriginTimestamp is then used in computing the phase offset (in ptp_message.cpp:972) and rate ratio (in ieee1588clock.cpp:309), thus resulting in incorrect values. The rate ratio is also subsequently used when processing the next FollowUp message, thus compounding the incorrect value.

To fix the issue, the correction value is computed by expressing the correctionField and path delay in the grandmaster time base as per the IEEE 802.1AS-2011 standard (clause 11.1.3). This correction field is applied to the preciseOriginTimestamp, which is then used to calculate the correct phase offset and rate ratio.

### DIFF
--- a/daemons/gptp/common/ptp_message.cpp
+++ b/daemons/gptp/common/ptp_message.cpp
@@ -408,6 +408,9 @@ PTPMessageCommon *buildPTPMessage
 		}
 		break;
 	case ANNOUNCE_MESSAGE:
+
+		XPTPD_INFO("*** Received Announce message");
+
 		{
 			PTPMessageAnnounce *annc = new PTPMessageAnnounce();
 			annc->messageType = messageType;
@@ -906,7 +909,7 @@ void PTPMessageFollowUp::processMessage(IEEE1588Port * port)
 	FrequencyRatio local_clock_adjustment;
 	FrequencyRatio local_system_freq_offset;
 	FrequencyRatio master_local_freq_offset;
-	int correction;
+	int64_t correction;
 	int32_t scaledLastGmFreqChange = 0;
 	scaledNs scaledLastGmPhaseChange;
 
@@ -959,9 +962,8 @@ void PTPMessageFollowUp::processMessage(IEEE1588Port * port)
 	master_local_freq_offset += 1.0;
 	master_local_freq_offset /= port->getPeerRateOffset();
 
-	correctionField = (uint64_t)
-		((correctionField >> 16)/master_local_freq_offset);
-	correction = (int) (delay + correctionField);
+	correctionField = (correctionField >> 16);
+	correction = (int64_t)((delay * master_local_freq_offset) + correctionField );
 
 	if( correction > 0 )
 	  TIMESTAMP_ADD_NS( preciseOriginTimestamp, correction );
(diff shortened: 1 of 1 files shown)


daemons/gptp/common/ptp_message.cpp
```diff
@@ -959,9 +962,8 @@ void PTPMessageFollowUp::processMessage(IEEE1588Port * port)
 	master_local_freq_offset += 1.0;
```
pinealservo: I have worked through the various ratios, and I believe this change is correct.

However, I think I found another existing bug in line 961; the equation from the specification is `cumulativeScaledRateOffset * 2^(-41)`, which means we should be dividing by `1ULL << 41` rather than `2ULL << 41`.

Since `FrequencyRatio` is `long double`, I guess we'll have enough precision leftover to deal with adding 1.0 (this effectively shifts 32 bits worth of mantissa precision to the right by 41 bits as we zero the exponent) but we'll still lose ~9 bits if 80-bit x86 extended precision floats are being used.  We could run into trouble in the future with these ratio calculations on systems where `long double` is the same thing as `double`, though; there are only effectively 53 bits of precision available there, so we'll lose 20 out of our original 32 bits.

andrew-elder: I must admit, I had to double check negative exponent rules. See https://en.wikipedia.org/wiki/Exponentiation#Negative_exponents. To my way of reading it
X * 2^(-41) == X / 2^ 41
so line 961
```
master_local_freq_offset /= 2ULL << 41;
```
should be correct. Am I missing something?

andrew-elder: Duh
2 ^ 41 != 2 << 41
eg
2 ^ 1 = 2
2 << 1 = 4
etc.
I get it now. Ignore my previous comment.

pinealservo: Exactly.  There is much subtlety required to do all these calculations correctly with minimal loss of precision.  We're not quite there yet, apparently, but this patch should definitely help.

k-gupta: Thanks for catching that. We were so engrossed in making sure correctionField value is added correctly we missed the values being used in the frequency calculation. Do you want me to make this change and add it to this pull request? 

Also a side note I just noticed that we send the pull request to the master branch. Do you want me to re initiate the pull request to the open-avb-next branch instead?

pinealservo: Yes, please make the pull request against open-avb-next.

k-gupta:
I have created a new pull request to open-avb-next with the changes suggested here.

[https://github.com/AVnu/Open-AVB/pull/447](url)

PR #446, opened 2016-07-28
Add -lrt to simple_rx build for Travis

pinealservo:
Travis is still running ancient Ubuntu systems as their default build image. In pre-2.17 versions of GLIBC, -lrt is required to use `clock_gettime`, and it doesn't seem to cause a problem in post-2.17 versions.


andrew-elder:
@pinealservo - thanks! Merging...

PR #447, opened 2016-07-29
Bug fixes for gPTP daemon

k-gupta:
This is an updated pull request for [https://github.com/AVnu/Open-AVB/pull/444](url) which incorporates the suggestions made by @pinealservo.


christopher-s-hall:
Correction field is signed (although I can't think of a good reason for it to be negative) and performing a right shift on a negative integer is "implementation defined". It's probably good enough either way, but the calculation would be more correct changing:

correctionField = (correctionField >> 16);

to

correctionField /= 1 << 16;

I get different results for correctionField = -65538. The first uses the floor the second truncates. I think truncation is probably the desired behavior.

pinealservo:
@christopher-s-hall Good point on the signed right-shift, although I think it's probably okay to put it off for a general review of numerical precision and portability concerns.

I'll give it a bit more time in case anyone else has further comments; I think it's ready for merge either as-is or with @christopher-s-hall's suggestion.

k-gupta:
I can make the change suggested by @christopher-s-hall

PR #450, opened 2016-08-08
Add support for "new" PTP cross-timestamp mechanism added in kernel version 4.6

christopher-s-hall:
Add support for HW timestamping & fix issue where link delay adjustments from .ini file are not applied correctly

### DIFF
--- a/daemons/gptp/linux/build/Makefile
+++ b/daemons/gptp/linux/build/Makefile
@@ -30,8 +30,18 @@
 
 ALTERNATE_LINUX_INCPATH=$(HOME)/header/include/
 
+# Check to see if PTP cross-timestamping is supported in hardware
+PRECISE_TIME_TEST = "\#include <linux/ptp_clock.h>\\n\
+	\#ifdef PTP_SYS_OFFSET_PRECISE\\nint main(){return 0;}\\n\#endif"
+HAS_PRECISE_TIME = $(shell echo $(PRECISE_TIME_TEST) | gcc -xc - \
+	-I$(ALTERNATE_LINUX_INCPATH) -o /dev/null > /dev/null 2>&1 ; echo $$?)
+
 CFLAGS_G = -Wall -std=c++0x -g -I. -I../../common -I../src \
 	-I$(ALTERNATE_LINUX_INCPATH)
+
+ifeq ($(HAS_PRECISE_TIME),0)
+	CFLAGS_G += -DPTP_HW_CROSSTSTAMP
+endif
 
 CPPFLAGS_G := $(CFLAGS_G) -Wnon-virtual-dtor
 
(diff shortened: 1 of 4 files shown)


andrew-elder:
How is PTP_SYS_OFFSET_PRECISE ever set? Should there be any comments anywhere describing how to set and use it?
Otherwise, changes look ok to me.

christopher-s-hall:
It will be defined in the kernel include file (include/uapi/linux) for kernel 4.6+ and not defined in previous versions. I’m fine with commenting it, but where?

Alternatively I could put something in the Makefile to test for this – probably a little messy – and add a wrapper. Now if someone were to grep for said wrapper it would be clear from the Makefile, in the form of a comment there, how it’s set.

All of this is somewhat academic because *eventually* these changes will be included as part of the standard headers and the need to wrap them in an ifdef will go away.

christopher-s-hall:
Absent feedback I opted for the second. The Makefile explicitly checks for support and sets an internal variable.

christopher-s-hall:
Added fix for issue #456

daemons/gptp/linux/build/Makefile
```diff
@@ -30,9 +30,19 @@
 
 ALTERNATE_LINUX_INCPATH=$(HOME)/header/include/
 
+# Check to see if PTP cross-timestamping is supported in hardware
+PRECISE_TIME_TEST = "\#include <linux/ptp_clock.h>\\n\
+	\#ifdef PTP_SYS_OFFSET_PRECISE\\nint main(){return 0;}\\n\#endif"
+HAS_PRECISE_TIME = $(shell echo $(PRECISE_TIME_TEST) | gcc -xc - \
+	-I$(ALTERNATE_LINUX_INCPATH) -o /dev/null > /dev/null 2>&1 ; echo $$?)
+
 CFLAGS_G = -Wall -std=c++0x -g -I. -I../../common -I../src \
```
andrew-elder: Wow, this is very clever. I personally think it is overkill in response to my "where does PTP_SYS_OFFSET" come from. I would have been 100% ok with a comment block in the .cpp module before the first usage of PTP_SYS_OFFSET_PRECISE explaining that kernels with precise time support have PTP_SYS_OFFSET_PRECISE defined in the linux/ptp_clock.h header file. My 2c. Others may have different opinions.

Sorry I was slow responding to the followup post on this topic.

christopher-s-hall:
The code is correct and in my opinion pretty clear. I would rather not spend more time on this. I don't think "overkill" should be a reason for code rejection.

andrew-elder:
@christopher-s-hall - acknowledged. Can anyone else sign off on this?

pinealservo:
How come this got closed?  I think there was a merge conflict indicated last time I saw it, but I didn't have time at the moment to see why.  Aside from that, it looked fine as far as I can tell.

christopher-s-hall:
Now pull #462. That has fewer commits (on second look, it didn't make a lot of sense to have so many commits), but is the same.

PR #459, opened 2016-08-18
Applicant state machine deviates from spec: txLA!/LO: stay in state LO

edgarholleis:


pinealservo:
Hi @edgarholleis, thanks for sending the pull request!

I've verified that this change brings the state machine code (for the state/event in question) into compliance with the state table, "Table 10-3--Applicant state table", of 802.1Q-2014.

It was probably a data entry error when the state machine was first coded, since it exists through the whole git history we have here, but I'm not certain. In order to figure out whether it might have been intentional, I've tried to work out what behavioral differences this change will produce.

Upon entering **LO**, a transmit opportunity should be requested. If granted for reasons other than the **LeaveAll** timer expiring, we should then get a **tx!** event rather than a **txLA!** event, in which case a transition to **VO** is what occurs after sending **Empty**.  In the cases where we transition to **LO** because we either sent or received a **LeaveAll**, our **LeaveAll** timer should be reset so there should be no problem getting a transmit opportunity before the next **LeaveAll** timer expiration.  However, **rLv** and **Redeclare!** events also cause a transition to **LO** from the various other Observer states, and it seems as if the **LeaveAll** timer could elapse before the transmit opportunity requested upon entering **LO** is granted, especially when a significant join timer is used for transmit opportunities.

If that occurs (which does seem rather unlikely) we would stop sending **Empty** in response to **LeaveAll** messages we received until we hear someone else send a **JoinEmpty** or **Empty** (which should happen as a result of our **LeaveAll** transmissions unless other weird circumstances occur, I think). Without this fix, however, we'd go back to **VO** in the weird case and respond to **LeaveAll** messages normally.

So I don't *think* this was an intentional deviation from the specification, and I think the circumstances under which this change would produce different behavior from before are rare enough that there's little risk in accepting the change. And of course it is the right thing to do according to the specification, even if it's not exactly obvious why the state machine in the specification is the way it is in this circumstance.

Unless someone else can argue for keeping the out-of-spec behavior (which does seem mostly benign, as far as out-of-spec behavior goes) I vote to accept the pull request.

intel-ethernet:
I agree with my esteemed college Levi that there is little risk in accepting the change. I don’t recall a specific reason to deviate from the spec, although I do not remember the leap where TxLa! on the LO state leads to VO per IEEE 802.1Q-2014 … by footnote of IEEE 802.1Q-2014:

“The VO, AO, and QO states represent states where the attribute is neither being declared by the Participant nor being registered
by any other station on the LAN. In implementations where dynamic creation and discarding of state machines is desirable, the
state machine can be discarded when in any of these states, pending a future requirement to declare or register that attribute value”

andrew-elder:
I agree with the previous two commenters. I'm going to accept this pull request and then rerun AVnu's test suite against the bug fix. I thought that I double checked all the state machine events and transitions in the distant past, but it looks like I missed one!

PR #464, opened 2016-08-31
Fixes to gPTP signaling message TLV values

kencarlino:
Starting with Wireshark 2.0.5 the signaling message in gPTP is reported as Malformed due to an incorrect value for the LengthField.

As defined in 802.1AS 2011 section 10.5.4.3.3 LengthField the length value shall be 12.
Currently gPTP is sending an incorrect value of 28.

Additionally for the signaling message TLV: 
Organization Subtype is 0x1 in the packet but it should be 0x2
LinkDelayInterval is 0x0 but for the Automotive Profile should be 127
Flags field is 0x0 in the packet but should be 0x3

### DIFF
--- a/daemons/gptp/common/avbts_message.hpp
+++ b/daemons/gptp/common/avbts_message.hpp
@@ -1059,16 +1059,16 @@ class SignallingTLV {
 	 */
 	SignallingTLV() {
 		tlvType = PLAT_htons(0x3);
-		lengthField = PLAT_htons(28);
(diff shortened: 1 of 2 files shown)


andrew-elder:
This change looks ok to me. Anyone else?

christopher-s-hall:
Looks like the right thing to me.

andrew-elder:
Thanks @christopher-s-hall, I'll go ahead and merge it.

Issue #460, opened 2016-08-23
gPTP: toByteString() function has no size checking


andrew-elder:
Current implementation blindly copies to target pointer without checking the size of the passed in memory area. Ideally there would be a size_t sizeof_byte_str_array parameter passed in by the caller so that the size could be checked before memcpy() type operations are undertaken.

jfornal:
I just created a Pull request for that issue (#468) - please review the changes and check whether this is exactly what you thought should be introduced into the code

pinealservo:
See my comment in #468 where I suggest that modifying `toByteString` might not be the best way to avoid blind copying. It's only blind if you don't know there's room before calling. It's basically a `memcpy` that automatically picks the right value for its length argument; I believe it's meant to have the same calling contract as `memcpy` itself and I think a better resolution would be to make note of the contract in the documentation and review callers to ensure there aren't any cases where the contract might be violated.  I don't think there are at present, though there are some places where assertions might be added to ensure protocol changes or different network interface types don't introduce regressions.

andrew-elder:
I'm going to close this issue for the moment. There really is no bug. If someone wants to open a new issue where parameters to `toByteString` are checked, they can.

PR #468, opened 2016-09-06
Open avb next - issue #460

jfornal:
This is a proposed solution for Andrew Elder's suggestion: changes made to "toByteString()" methods in gptp/common/avbts_message.hpp and gptp/common/ptp_message.cpp:

a) Added a sizeof_byte_str_array parameter to all three "toByteString" methods (SignallingTLV and FollowupTLV are used, scaledNS has been altered for future reference and consistency purposes)

b) A separate "available_size" variable is being created before each of the method invocation . It is being initialized with space left available in the buffer (buf_t) to which the data is being copied further on

### DIFF
--- a/daemons/gptp/common/ptp_message.cpp
+++ b/daemons/gptp/common/ptp_message.cpp
@@ -899,8 +899,11 @@ void PTPMessageFollowUp::sendPort(IEEE1588Port * port,
 	tlv.setGMTimeBaseIndicator(tbi_NO);
 
 	size_t available_size = sizeof(buf_t) - (PTP_COMMON_HDR_LENGTH + PTP_FOLLOWUP_LENGTH
-						+port->getPayloadOffset());
-	tlv.toByteString(buf_ptr + PTP_COMMON_HDR_LENGTH + PTP_FOLLOWUP_LENGTH, available_size);
+						+ port->getPayloadOffset());
+	int result = tlv.toByteString(buf_ptr + PTP_COMMON_HDR_LENGTH + PTP_FOLLOWUP_LENGTH,
+						 available_size);
+	if(result != 0)
+		GPTP_LOG_ERROR("toByteString returned an error - invalid size passed");
 
 	GPTP_LOG_VERBOSE
 		("Follow-Up Time: %u seconds(hi)", preciseOriginTimestamp.seconds_ms);
@@ -1767,7 +1770,10 @@ void PTPMessageSignalling::sendPort(IEEE1588Port * port, PortIdentity * destIden
 
 	size_t available_size = sizeof(buf_t) - (PTP_COMMON_HDR_LENGTH + PTP_SIGNALLING_LENGTH
 						+ port->getPayloadOffset());
-	tlv.toByteString(buf_ptr + PTP_COMMON_HDR_LENGTH + PTP_SIGNALLING_LENGTH, available_size);
+	int result = tlv.toByteString(buf_ptr + PTP_COMMON_HDR_LENGTH + PTP_SIGNALLING_LENGTH,
+						 available_size);
+	if(result != 0)
+		GPTP_LOG_ERROR("toByteString returned an error - invalid size passed");
 
 	port->sendGeneralPort(PTP_ETHERTYPE, buf_t, messageLength, MCAST_OTHER, destIdentity);
 }
--- a/daemons/gptp/common/ptp_message.cpp
+++ b/daemons/gptp/common/ptp_message.cpp
@@ -774,8 +774,8 @@ void PTPMessageAnnounce::sendPort(IEEE1588Port * port,
 	       &stepsRemoved_l, sizeof(stepsRemoved));
 	memcpy(buf_ptr + PTP_ANNOUNCE_TIME_SOURCE(PTP_ANNOUNCE_OFFSET),
 	       &timeSource, sizeof(timeSource));
+	
 	tlv.toByteString(buf_ptr + PTP_COMMON_HDR_LENGTH + PTP_ANNOUNCE_LENGTH);
-
 	port->sendGeneralPort(PTP_ETHERTYPE, buf_t, messageLength, MCAST_OTHER, destIdentity);
 	port->incCounter_ieee8021AsPortStatTxAnnounce();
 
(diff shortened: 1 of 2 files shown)


daemons/gptp/common/ptp_message.cpp
```diff
@@ -897,7 +897,10 @@ void PTPMessageFollowUp::sendPort(IEEE1588Port * port,
 	/*Change time base indicator to Network Order before sending it*/
 	uint16_t tbi_NO = PLAT_htonl(tlv.getGMTimeBaseIndicator());
 	tlv.setGMTimeBaseIndicator(tbi_NO);
-	tlv.toByteString(buf_ptr + PTP_COMMON_HDR_LENGTH + PTP_FOLLOWUP_LENGTH);
+
+	size_t available_size = sizeof(buf_t) - (PTP_COMMON_HDR_LENGTH + PTP_FOLLOWUP_LENGTH
+						+port->getPayloadOffset());
```
andrew-elder: There should be a space after the "+"

andrew-elder:
The changes basically look ok to me. If we are touching this code, it would also be good to return an int so that there are no silent (hidden) failures. I think the convention (in this gptp implementation) is to return 0 on success and -1 on failure. Then, in the caller, the return would be checked and an error logged if the return code is non-zero.

jfornal:
I added the things you mentioned: do you have any other proposals/comments? Please let me know : )

daemons/gptp/common/ptp_message.cpp
```diff
@@ -1762,7 +1768,12 @@ void PTPMessageSignalling::sendPort(IEEE1588Port * port, PortIdentity * destIden
...
 
-	tlv.toByteString(buf_ptr + PTP_COMMON_HDR_LENGTH + PTP_SIGNALLING_LENGTH);
+	size_t available_size = sizeof(buf_t) - (PTP_COMMON_HDR_LENGTH + PTP_SIGNALLING_LENGTH
+						+ port->getPayloadOffset());
+	int result = tlv.toByteString(buf_ptr + PTP_COMMON_HDR_LENGTH + PTP_SIGNALLING_LENGTH,
+						 available_size);
+	if(result != 0)
+		GPTP_LOG_ERROR("toByteString returned an error - invalid size passed");
 
 	port->sendGeneralPort(PTP_ETHERTYPE, buf_t, messageLength, MCAST_OTHER, destIdentity);
```
pinealservo: This got to be rather lengthy.  I want to start out saying that I appreciate the effort and initiative in making the pull request and this only got as long as it did because of underlying stuff I want to have a conversation about, not because I want to be critical of the code in the pull request!  It's a great implementation (with a little bit more needed) of the requested change, but as I looked deeper I started to question some assumptions behind it.

**Regarding the pull request as-is**

Do we still want to send (here and in the other `sendPort` method) if the `tlv.toByteString` calls fail?  The headers at this point include the length of the tlv, but we haven't written any tlv data, and we couldn't write it because there wasn't enough room, so the length in our header exceeds the size of the actual payload buffer. We also pass `messageLength` (which includes the tlv length) to our port's send routine, so it will happily read beyond `buf_t` (which as a side note is a poor name for a variable; POSIX reserves all names ending in `_t`, and the ones it uses are for *types*, not variables, so it's also confusing) into other parts of the stack.

**Thoughts on the general approach**

If I understand correctly, the only real non-static length we're dealing with here is `port->getPayloadOffset()`, and the return value of that method should never change for the lifetime of the port. Ideally we'd make a static assertion that our static buffer size is big enough manage the largest possible getPayloadOffset() value for all the different port types, but I don't think there's a way to express that sort of thing directly in C++.

Second best would be a dynamic assertion along with a unit test that combines calls to this with all the possible port types.  But we're definitely not there with the unit test infrastructure on gPTP yet.

In any case, I'm not sure adding the length checking to `toByteString` is the right thing to do.  Certainly length checking *does* need to be done, but a note in the documentation for `toByteString` that it assumes the caller has already done length checking would be sufficient. It is often useful for performance reasons to have low-level stream writing primitives that don't do checks like that so that you can do one higher-level check followed by a fixed-length series of the lower-level writer calls that don't check. If you check in every primitive, all the checks can add up to a significant portion of packet construction time, and most of them are redundant. It is something of a performance vs. mistake-proofing trade-off, but the code generally seems to follow this approach and it would take a lot of rework to move it to a more 'mistake-proof' approach. As the relevant code is mostly tested and working, I don't see much value in doing that.

In these TLV cases, a single length check per sendPort method should be sufficient to guarantee safety, and I think it's a great thing to add to avoid regressions in the future, though it should probably be done via some sort of assertion or check-and-abort near the beginning of the methods.

andrew-elder:
I am wondering where all the above discussion leaves us?

@pinealservo - I see your point that we shouldn't have to check the size of the target buffer because we "know" the size is correct due to all the code that calls toByteString() being part of the same codebase.

We could add a #define and a few static_assert() operations to catch issues at compile time. Something along the lines of

    #define MAX_PAYLOAD_OFFSET 16

Then in the implementations of `getPayloadOffset()` add

    static_assert(offset <= MAX_PAYLOAD_OFFSET)

and just before the calls to `toByteString()` another `static_assert()` would be added using `MAX_PAYLOAD_OFFSET` to represent the worst case return from `getPayloadOffset()`.

It's not 100% perfect because someone could implement a version of `getPayloadOffset()` without including the check against `MAX_PAYLOAD_OFFSET`.

Comments?

pinealservo:
The easiest thing to do here is nothing; I don't think there's actually an existing bug that this is required to fix, is there? 

I also have no objections to just adding a check after`getPayloadOffset` to ensure there's enough room for the `memcpy` and `tlv->toByteString` calls to succeed.  If that's done, we should ensure that we bail without sending the packet if the size check fails.

And if you would prefer to have checks in `toByteString` anyway, that's fine too; we just need to also ensure in this case that we don't try to send a packet for which we haven't allocated enough buffer space for the length we're telling the transmitter we're giving it.  That's basically this pull request plus early exits when `toByteString` fails.

Or we could try to do something fancier that is safe and also low-overhead.  I wouldn't oppose that, but I don't have time to commit to it either.

andrew-elder:
I'm going to close this pull request as it no longer seems like the (obviously) correct path forward. @jfornal has agreed to this (out-of-thread). I'm also going to close #460.

Issue #436, opened 2016-07-20
gPTP: timestamping failures on link up/down events


merrilldav:
While running gPTPd on Linux from the master branch (compiled with generic architecture), we are seeing timestamping errors after a network port is reset using 'ifconfig ethx down' and 'ifconfig ethx up'.  We think we found a fix, but it is not covering an important edge case, as I'll describe below.

The initial error we saw:
`ERROR at 731 in ../../common/ieee1588port.cpp: Error (TX) timestamping PDelay request, error=-72
Invalid TX`

gPTP never recovers and remains in master mode.  After killing the gPTPd daemon and restarting, it functions again.  This appears to relate to issue #254, which added fa7a482 to detect link state changes and call [timestamper_init()](https://github.com/AVnu/Open-AVB/blob/32dab06dd89427114d26be995ffdbf45289adb4e/daemons/gptp/common/ieee1588port.cpp#L588-L591) when LINKUP occurs.  However, in our testing we found that this change did not address our problem.

After some investigation, we discovered that the root cause is that the RX listener thread exits when it detects a failure receiving a packet, ensuring that no further packets will be received.  Removing the 'break' statement in the following code appears to fix the problem: [openPort()](https://github.com/AVnu/Open-AVB/blob/32dab06dd89427114d26be995ffdbf45289adb4e/daemons/gptp/common/ieee1588port.cpp#L354-L357)

However, we further found that the call to timestamper_init() causes several problems:
  1) an apparent deadlock shortly after LINKUP
  2) a ~40 byte memory leak each time the network cable is unplugged, seen in Valgrind
  3) adding redundant copies of the LinuxNetworkInterface object to iface_list in [HWTimestamper_init()](https://github.com/AVnu/Open-AVB/blob/32dab06dd89427114d26be995ffdbf45289adb4e/daemons/gptp/linux/src/linux_hal_generic.cpp#L252-L255)

We removed the call to timestamper_init(), and this appeared to make everything work.  However, after further testing running ifconfig down/up in a loop on the slave, we found that the _master_ would sometimes lose the ability to timestamp any TX packets:

`ERROR at 729 in ../../common/ieee1588port.cpp: Error (TX) timestamping PDelay request, error=-72
ERROR at 1125 in ../../common/ptp_message.cpp: Error (TX) timestamping PDelay Response (Retrying-0), error=-72
ERROR at 1137 in ../../common/ptp_message.cpp: Error (TX) timestamping PDelay Response, error=-72
[...]

PR #469, opened 2016-09-07
Change link up handling of hardware timestamper

pinealservo:
As per issue #436 there are some problems with the way we were calling `timestamper_init` when we received link up events. That function is only designed to be called once, and causes resource leaks when called multiple times.

These patches create a 'reset' interface to the timestamper and call that instead on link up events. For the moment, it only resets the frequency adjustment, but if there are other useful things that need to happen when the link is restored we'll have a safe place to add them.

### DIFF
--- a/daemons/gptp/common/avbts_port.hpp
+++ b/daemons/gptp/common/avbts_port.hpp
@@ -1139,6 +1139,12 @@ class IEEE1588Port {
 	void timestamper_init(void);
 
 	/**
+	 * @brief Resets the hwtimestamper
+	 */
+	void timestamper_reset(void);
+
+
+	/**
 	 * @brief  Gets RX timestamp based on port identity
 	 * @param  sourcePortIdentity [in] Source port identity
 	 * @param  sequenceId Sequence ID
--- a/daemons/gptp/common/ieee1588.hpp
+++ b/daemons/gptp/common/ieee1588.hpp
@@ -478,6 +478,13 @@ public:
 		{ return true; }
 
 	/**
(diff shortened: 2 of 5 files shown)


andrew-elder:
Code looks good to me. I'll fix the Windows build after this pull request has been completed. My only comment is that I think it helps commit scanning at a later date if we put something like "gPTP " at the start of the commit message.

pinealservo:
Hmm, I didn't realize I'd broken the Windows build.  I can take a look at that and redo the commit messages while we wait for someone else to take a look at it.  Maybe @christopher-s-hall or @davidcemin or @merrilldav have some thoughts?

andrew-elder:
Instead of getting on an Avnu conference call, I'm hoping some others can respond to @pinealservo's pull request.

andrew-elder:
I'm going to "eat crow" as they say. I assumed the Windows build was broken from reading changes to the HAL module interface, but only the derived linux class has the new method so all is fine. I have just built the Windows version and it compiles correctly. If one other person can please add a "this looks ok to me" comment I'll merge (pending commit comment rewording).

christopher-s-hall:
I don't see any problem if it doesn't break the Windows build.

andrew-elder:
@christopher-s-hall - thanks!

pinealservo:
I just finally got a Windows build environment working again and the Windows build does indeed work. I'll work on the commit rewording now.

pinealservo:
OK, I got the commit messages updated with gPTP prepended.

PR #471, opened 2016-09-07
gPTP - add some NULL  pointer checking

andrew-elder:
This pull request adds some NULL pointer checking related to announce processing.

### DIFF
--- a/daemons/gptp/common/ieee1588port.cpp
+++ b/daemons/gptp/common/ieee1588port.cpp
@@ -618,7 +618,7 @@ void IEEE1588Port::processEvent(Event e)
 					}
 					if (EBest == NULL) {
 						EBest = ports[j]->calculateERBest();
-					}else if (ports[j]->calculateERBest()) {
+					} else if (ports[j]->calculateERBest()) {
 						if (ports[j]->calculateERBest()->isBetterThan(EBest)) {
 							EBest = ports[j]->calculateERBest();
 						}
--- a/daemons/gptp/common/ieee1588port.cpp
+++ b/daemons/gptp/common/ieee1588port.cpp
@@ -598,8 +598,7 @@ void IEEE1588Port::processEvent(Event e)
 
 	case STATE_CHANGE_EVENT:
 		if (!automotive_profile) {       // BMCA is not active with Automotive Profile
-			if ( clock->getPriority1() != 255 && qualified_announce != NULL ) {
-
+			if ( clock->getPriority1() != 255 ) {
 				int number_ports, j;
 				PTPMessageAnnounce *EBest = NULL;
 				char EBestClockIdentity[PTP_CLOCK_IDENTITY_LENGTH];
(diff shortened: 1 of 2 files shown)


daemons/gptp/common/ieee1588port.cpp
```diff
@@ -598,7 +598,8 @@ void IEEE1588Port::processEvent(Event e)
 
 	case STATE_CHANGE_EVENT:
 		if (!automotive_profile) {       // BMCA is not active with Automotive Profile
-			if ( clock->getPriority1() != 255 ) {
+			if ( clock->getPriority1() != 255 && qualified_announce != NULL ) {
```
pinealservo: I think checking qualified_announce here is redundant;  it's set right before STATE_CHANGE_EVENT is queued for this port and it's also checked below when `calculateERBest()` is called on each port in your addition below.  I can't see that it's accessed between here and the check below.

daemons/gptp/common/ieee1588port.cpp
```diff
@@ -617,13 +618,17 @@ void IEEE1588Port::processEvent(Event e)
 					}
 					if (EBest == NULL) {
 						EBest = ports[j]->calculateERBest();
```
pinealservo: This could still be `NULL` afterward if the port hasn't received an announce yet and `calculateERBest()` returns `NULL` at this point.

daemons/gptp/common/ieee1588port.cpp
```diff
@@ -617,13 +618,17 @@ void IEEE1588Port::processEvent(Event e)
...
-					} else {
+					}else if (ports[j]->calculateERBest()) {
```
pinealservo: Would be nice to not remove the space between `}` and `else`.

pinealservo:
I think the logic covers all the possibilities that could introduce badness in a multi-port implementation.  I thought there was a missing one and made a comment about it, but I was mistaken and deleted it; if you get an email and can't find the reference, that's why.

If you'd like to leave the possibly-redundant check, I don't think it hurts anything and I may have missed the reason for it, so I won't object.

daemons/gptp/common/ieee1588port.cpp
christopher-s-hall: Agree with @pinealservo there's not really any way I could see that qualified_announce would be NULL. If you want to check for NULL, I think it would be better to check the value of EBest, but still not necessary.

andrew-elder: This line was added due to null pointer faults (gPTP crashes) we observed in-house. I see that EBest checks were added here on the same day, so perhaps additional unnecessary checks were added. I'll update with this commit removed.

andrew-elder:
I think I have now addressed all comments made by @pinealservo and @christopher-s-hall

pinealservo:
Looks good to me!

PR #477, opened 2016-09-13
Fixes to Makefiles and compiler warnings and errors

### LINKED ISSUE #473: Array subscript is above array bounds warning in maap_linux.c
I am adding -O2 to Open AVB daemons Makefile CFLAGS and one of the warning shown is:

```
maap_linux.c: In function ‘main’:
maap_linux.c:198:16: warning: array subscript is above array bounds [-Warray-bounds]
  seed = src_mac[6] + time(NULL);
                ^
```

I have checked the source code daemons/maap/linux/maap_linux.c, in function main(), src_mac is declared as 6 bytes and seed is an unsigned long.
```
uint32_t seed;
uint8_t src_mac[6];
```

May I know if it is a mistake or it is by purpose to use src_mac[6] to assign the seed?

Regards,
Roland Hii

rhii:
Since added simple_rx, the clean target does not properly clean all the generated object and executable files. The changes here corrected the corresponding Makefiles to be able to remove all the generated files.

Besides that, when added -O2 to Open AVB daemons Makefile CFLAGS, there are some compile warnings and errors shown. The changes here also fix those warnings and errors.

This should fix #473 too.

Regards,
Roland Hii

### DIFF
--- a/daemons/gptp/linux/src/linux_hal_persist_file.cpp
+++ b/daemons/gptp/linux/src/linux_hal_persist_file.cpp
@@ -123,7 +123,9 @@ public:
 
 		bool result = false;
 		if (memoryDataLength > storedDataLength) {
-			ftruncate(persistFD, memoryDataLength);
+			int ret = ftruncate(persistFD, memoryDataLength);
+			if (ret != 0) {
+			}
 			if (restoredata != ((void *)-1)) {
 				restoredata = mremap(restoredata, storedDataLength, memoryDataLength, MREMAP_MAYMOVE);
 			}
--- a/daemons/gptp/linux/src/linux_hal_persist_file.cpp
+++ b/daemons/gptp/linux/src/linux_hal_persist_file.cpp
@@ -125,6 +125,7 @@ public:
 		if (memoryDataLength > storedDataLength) {
 			int ret = ftruncate(persistFD, memoryDataLength);
 			if (ret != 0) {
+				printf("Failed to extend stored data length from %ld to %ld, %s\n", storedDataLength, memoryDataLength, strerror(errno));
 			}
 			if (restoredata != ((void *)-1)) {
 				restoredata = mremap(restoredata, storedDataLength, memoryDataLength, MREMAP_MAYMOVE);
--- a/examples/live_stream/listener.c
+++ b/examples/live_stream/listener.c
@@ -198,8 +198,10 @@ int main(int argc, char *argv[ ])
 			fprintf(stderr,"frame sequence = %lld\n", frame_sequence++);
 			h1722 = (seventeen22_header *)((uint8_t*)frame + sizeof(eth_header));
 			length = ntohs(h1722->length) - sizeof(six1883_header);
-			write(1, (uint8_t *)((uint8_t*)frame + sizeof(eth_header) + sizeof(seventeen22_header) +
+			rc = write(1, (uint8_t *)((uint8_t*)frame + sizeof(eth_header) + sizeof(seventeen22_header) +
 				sizeof(six1883_header)), length);
+			if (rc != length) {
+			}
 		} else {
 			fprintf(stderr,"recvfrom() error for frame sequence = %lld\n", frame_sequence++);
 		}
--- a/examples/live_stream/listener.c
+++ b/examples/live_stream/listener.c
@@ -200,7 +200,8 @@ int main(int argc, char *argv[ ])
 			length = ntohs(h1722->length) - sizeof(six1883_header);
 			rc = write(1, (uint8_t *)((uint8_t*)frame + sizeof(eth_header) + sizeof(seventeen22_header) +
 				sizeof(six1883_header)), length);
-			if (rc != length) {
+			if (rc == -1) {
+				fprintf(stderr, "Failed to write %d bytes: %s (%d)\n", length, strerror(errno), errno);
 			}
 		} else {
 			fprintf(stderr,"recvfrom() error for frame sequence = %lld\n", frame_sequence++);
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ simple_listener:
 	$(call descend,examples/$@)
 
 simple_listener_clean:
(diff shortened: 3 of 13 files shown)


andrew-elder:
Thanks for this pull request @rhii. I took a quick look and I wonder if you would be willing to split the pull request in to 2 separate requests? One with makefile changes only and a second one with the compiler warnings fixes. I ask because my initial read through indicates that the makefile changes look to be trivial, but I suspect some of the warnings updates will generate discussion. Particularly the meaningless handling of return codes that was added in a couple of spots.

daemons/gptp/linux/src/linux_hal_persist_file.cpp
```diff
@@ -123,7 +123,9 @@ class LinuxGPTPPersistFile : public GPTPPersist {
 
 		bool result = false;
 		if (memoryDataLength > storedDataLength) {
-			ftruncate(persistFD, memoryDataLength);
+			int ret = ftruncate(persistFD, memoryDataLength);
```
pinealservo: I don't understand the point of this.  It doesn't change any behavior, so the compiler will most likely not even emit any code for it. I'm all for adding return code checks, but they should at least use the existing log mechanisms to indicate what happened when a non-success return code is found.

examples/live_stream/listener.c
```diff
@@ -198,8 +198,10 @@ int main(int argc, char *argv[ ])
 			fprintf(stderr,"frame sequence = %lld\n", frame_sequence++);
 			h1722 = (seventeen22_header *)((uint8_t*)frame + sizeof(eth_header));
 			length = ntohs(h1722->length) - sizeof(six1883_header);
-			write(1, (uint8_t *)((uint8_t*)frame + sizeof(eth_header) + sizeof(seventeen22_header) +
+			rc = write(1, (uint8_t *)((uint8_t*)frame + sizeof(eth_header) + sizeof(seventeen22_header) +
```
pinealservo: See previous comment about checking return codes.

pinealservo:
Most of these look like good fixes, but I am dubious about adding empty conditional blocks for the unchecked return value warnings. We can always add this as-is and put something reasonable in as a separate pull request, though.  Don't want to block it over that, since there should be no runtime impact.

I am also unclear about why you got errors regarding making a shared object. I don't recall ever seeing those errors or seeing any shared object build rules in the system. I guess we'd need the -fPIC options if we wanted a shared library, but I don't see that you've added any shared library builds to the system. There is potentially some runtime overhead to PIC, so I'd like to at least hear a justification for it, although I don't have a problem with it if there's a good reason.

daemons/gptp/linux/src/linux_hal_persist_file.cpp
rhii: Thanks for your comment @pinealservo. I have added a new commit to print a message when a non-success code is returned.

Regards,
Roland Hii

examples/live_stream/listener.c
rhii: Thanks for your comment @pinealservo. I have added a new commit to print a message when a non-success code is returned.

Regards,
Roland Hii

rhii:
Thanks to your comments, Andrew and Levi.

@andrew-elder I am fine to split the pull request into 2. I will do that if there are more discussion on the warnings update.

@pinealservo The empty conditional blocks do eliminate the compiler warnings on my side. I have also added a new commit to print a message when a non-success code is returned.

Regarding the shared object, you are right that we need the -fPIC option. I do not add the -fPIC option to the Makefile. I am using bitbake build system to build the project and somehow the system added that -fPIC option recently and prevented me to build this project successfully. This triggers me to create this pull request as I think this would probably benefit others as well.

Please let me know if this "shared object" is not desired and I will remove it from the pull request.

Thank you.

Regards,
Roland Hii

andrew-elder:
@rhii - don't bother splitting the pull request now. There has been so much discussion here already that we may as well continue as we are. It is something to keep in mind for the future though.

rhii:
@andrew-elder - I still use printf() in linux_hal_persist_file.cpp because printf() is still being used in that file, not converted to use log macro yet. If needed I can submit another pull request to change the printf() to log macro, but I don't want to mix it in this pull request.

Noted on the splitting of pull request. I will keep this in mind for future requests.

Regards,
Roland Hii

andrew-elder:
@rhii - fair enough. And I'm glad you felt free to disagree with me! I must admit I didn't review the entire file before making my comment.
Regards,
Andrew

pinealservo:
@rhii Is there a repository somewhere for the .bb file that you're using to build the project, or is it a private project?  It may be more appropriate to adjust it to not attempt to build shared libraries than to add -fPIC to our static library build makefiles.

rhii:
@pinealservo - it is not my .bb file that adds the shared object compile option. This error
```
error: obj/ini.o: relocation R_X86_64_32 against `.rodata' can not be
 used when making a shared object; recompile with -fPIC
```
is shown after I updated my build system and I do not know where is the shared object compile option being added.

If it is a concern to add the -fPIC flag to Open AVB makefiles, I can remove those -fPIC flag related changes and submit a new patch. Please let me know.

Thank you for your comments.

Regards,
Roland Hii

pinealservo:
@rhii Which set of bitbake rules are you using that has a build for OpenAVB? I realize your build is broken by this and it's not something you did yourself, but it's technically a bug in the bitbake configuration rather than OpenAVB. I'd like to see if I can figure out why it's doing what it's doing and submit a bug report or possibly a patch to that set of rules. If they're a private set of bitbake rules you don't have any influence over, maybe I can help you come up with a local override .bb file that fixes it.

For now, please remove the -fPIC changes; if you'd like, you can submit them as a separate pull request. I'd like input from some others on whether it's a good idea to add that flag to the build or not. I think the rest of your changes can be merged as-is, though.

andrew-elder:
@rhii - thanks for the updated pull request. I think that is the last small item of this pull request that was under review, so I'm going to go ahead and merge the pull request.

PR #478, opened 2016-09-13
gPTP: fix MSVC compiler warnings

andrew-elder:
Minor typecast additions to correct Microsoft compiler warnings.


pinealservo:
Looks reasonable to me.

Issue #448, opened 2016-08-05
ISO C++98 does not support the ‘hh’ gnu_printf length modifier [-Werror=format=] add("%02hhx ", byteMemory[currentPos + p]);


amustafa:
When running:

    mkdir tmp
    cd tmp
    cmake .. -G "Unix Makefiles"
    make

as instructed on the main page, I get:

    [ 11%] Building CXX object thirdparty/cpputest/src/CppUTest/CMakeFiles/CppUTest.dir/MemoryLeakDetector.cpp.o
    /home/amustafa/workspace/Open-AVB/thirdparty/cpputest/src/CppUTest/MemoryLeakDetector.cpp: In member function ‘void SimpleStringBuffer::addMemoryDump(const void*, size_t)’:
    /home/amustafa/workspace/Open-AVB/thirdparty/cpputest/src/CppUTest/MemoryLeakDetector.cpp:75:54: error: ISO C++98 does not support the ‘hh’ gnu_printf length modifier [-Werror=format=]
             add("%02hhx ", byteMemory[currentPos + p]);
                                                      ^
    cc1plus: all warnings being treated as errors
    thirdparty/cpputest/src/CppUTest/CMakeFiles/CppUTest.dir/build.make:326: recipe for target  'thirdparty/cpputest/src/CppUTest/CMakeFiles/CppUTest.dir/MemoryLeakDetector.cpp.o' failed
    make[2]: *** [thirdparty/cpputest/src/CppUTest/CMakeFiles/CppUTest.dir/MemoryLeakDetector.cpp.o] Error 1
    CMakeFiles/Makefile2:1033: recipe for target 'thirdparty/cpputest/src/CppUTest/CMakeFiles/CppUTest.dir/all' failed
    make[1]: *** [thirdparty/cpputest/src/CppUTest/CMakeFiles/CppUTest.dir/all] Error 2
    Makefile:138: recipe for target 'all' failed
    make: *** [all] Error 2

I was able to get around this by opening the tmp/CMaleCache.txt file and changing the following flag:

       //Compile with C++11 support
      -  C++11:BOOL=OFF
      +  C++11:BOOL=ON

t-hofmann:
simple change in
Open-AVB/thirdparty/cpputest/src/CppUTest/MemoryLeakDetector.cpp
line 75
from
         add("%02hhx ", byteMemory[currentPos + p]);
to
         add("%02hx ", byteMemory[currentPos + p]);

[...]

PR #485, opened 2016-09-22
Enable C++11 support in CppUTest

pinealservo:
This is a fix for #448 

It enables C++11 support in the CppUTest build via a change to the top-level CMakeLists.txt file, and also fixes some warnings from our MRPD unit tests that showed up as a result of the change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,17 @@ git:
 before_install:
   - sed -i 's/git@github.com:/https:\/\/github.com\//' .gitmodules
   - git submodule update --init --recursive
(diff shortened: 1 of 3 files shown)


pinealservo:
Well that's wonderful.  I guess I'll leave this open while I look for a way to make Travis happy.

pinealservo:
OK; now this pull request does what it says above, plus it upgrades the GCC suite in the Travis config to version 4.8, which is new enough to support the c++11 standard via the command line flag.

andrew-elder:
Thanks @pinealservo . All looks good to me. Unless I see any comments that trigger further discussion I'll merge this 24 hours from now.

Issue #465, opened 2016-09-01
gPTP - SYNC_INTERVAL_TIMEOUT_EXPIRES automotive sync interval seems incorrect


andrew-elder:
This code seems to be incorrect to me:

https://github.com/AVnu/Open-AVB/blob/open-avb-next/daemons/gptp/common/ieee1588port.cpp#L1099

The comment talks of 16 millisecond timer in the if (!automotive_profile) section of the code.

What do the automotive experts think?

andrew-elder:
Yes, but the accelerated logic seems to be selected by the *if (!automotive_profile)* code path - notice the  "!".

andrew-elder:
Does a wireshark trace show the accelerated sync path works?

andrew-elder:
Thanks for the comment @kencarlino.

christopher-s-hall:
The general idea was to fire off several (I think there was a parameter for this) sync messages in quick succession.  I may have changed the number to make the comment not match. The pdelay wasn't supposed to start (using cached values) until all of these "fast" syncs were sent.

I don't know of anyone using this code and it can probably be removed since it doesn't comply with any standards.  It is also true that the code implementing the automotive profile renders this code mostly useless.  I say dump it.

andrew-elder:
Thanks for the comments @christopher-s-hall  and @kencarlino. Just to be sure I understand what is being said here, I'm going to restate what I understand is being said and proposed code changes.

The main point I take from comments so far is that there is no longer a requirement to fire off some number of sync messages in quick succession. That means I can remove *accelerated_sync_count* from here
https://github.com/AVnu/Open-AVB/blob/0c8a455ec1d84ff09608dcd74cfe5726da59385a/daemons/gptp/common/avbts_port.hpp#L223
and *_accelerated_sync_count* from
https://github.com/AVnu/Open-AVB/blob/0c8a455ec1d84ff09608dcd74cfe5726da59385a/daemons/gptp/common/avbts_port.hpp#L322
and clean up any associated interfaces/references.

And then this if() else sequence 
https://github.com/AVnu/Open-AVB/blob/0c8a455ec1d84ff09608dcd74cfe5726da59385a/daemons/gptp/common/ieee1588port.cpp#L1099
collapses to
```
		syncDone();
		wait_time = ((long long) (pow((double)2, getSyncInterval()) * 1000000000.0));
[...]

PR #486, opened 2016-09-22
gPTP: remove accelerated SYNC send on startup. See discussion in #465.

andrew-elder:
Completely remove accelerated sync on gPTP daemon startup. I think I have covered everything - comments welcomed.

### DIFF
--- a/daemons/gptp/common/ieee1588port.cpp
+++ b/daemons/gptp/common/ieee1588port.cpp
@@ -140,7 +140,6 @@ IEEE1588Port::IEEE1588Port(IEEE1588PortInit_t *portInit)
 
 	/*TODO: Add intervals below to a config interface*/
 	log_mean_sync_interval = initialLogSyncInterval;
-	_accelerated_sync_count = portInit->accelerated_sync_count;
 	log_mean_announce_interval = 0;
 	log_min_mean_pdelay_req_interval = initialLogPdelayReqInterval;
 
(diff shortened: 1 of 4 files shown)


andrew-elder:
I'm hoping @christopher-s-hall can take a quick look at this. I not 100% sure I have implemented his suggestion as discussed in issue #465. Let me know and I'll make the required/suggested changes. Anyone else have any comments?

daemons/gptp/common/ieee1588port.cpp
```diff
@@ -1104,36 +1099,11 @@ void IEEE1588Port::processEvent(Event e)
...
-					    wait_time;
-					wait_time = wait_time > EVENT_TIMER_GRANULARITY ? wait_time :
-					            EVENT_TIMER_GRANULARITY;
-					startSyncIntervalTimer(wait_time);
-				}
-			}
-			else {
-				// Automotive Profile
-				syncDone();
+			syncDone();
```
pinealservo: The previous code subtracted the old `wait_time` from the new `wait_time`; as far as I can tell this sets the wait interval from the beginning of the Sync interval rather than from the point we've finished waiting for the hardware to give us a TX Timestamp for our first Sync.

I'm not sure whether you left this out intentionally to simplify things or if you missed it.  I'm not sure if it makes much difference correctness-wise.

andrew-elder: I did see the wait time adjustment math. I concluded it was most relevant to the automotive fast sync case where TX timestamp read time might have a large effect percentage wise on the Sync interval timing. I didn't think removing the adjustment would have any negative operational effect - although I must admit I haven't hooked up test equipment to try to figure out if there is an such effect.

pinealservo:
Any comment @christopher-s-hall, or anyone else? We need another reviewer on this.

daemons/gptp/common/ieee1588port.cpp
christopher-s-hall: It looks fine. I'm not too worried about the interval adjustment at this point.

pinealservo:
Thanks, @christopher-s-hall

PR #489, opened 2016-10-05
gPTP: IEEE1588Port::processEvent(), start and stop pDelay on LINKUP/D…

andrew-elder:
…OWN events

Without stopping pDelay timers, continuous log output is generated when
the network cable is disconnected.

### DIFF
--- a/daemons/gptp/common/ieee1588port.cpp
+++ b/daemons/gptp/common/ieee1588port.cpp
@@ -712,6 +712,8 @@ void IEEE1588Port::processEvent(Event e)
 		break;
 
 	case LINKUP:
+		haltPdelay(false);
+		startPDelay();
 		if (automotive_profile) {
 			GPTP_LOG_EXCEPTION("LINKUP");
 		}
@@ -771,6 +773,8 @@ void IEEE1588Port::processEvent(Event e)
 		break;
 
 	case LINKDOWN:
+		setAsCapable(false);
+		stopPDelay();
 		if (automotive_profile) {
 			GPTP_LOG_EXCEPTION("LINK DOWN");
 		}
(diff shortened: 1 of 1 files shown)


andrew-elder:
Plan to merge this after pull request from open-avb-next to master has been completed.

andrew-elder:
Should the automotive profile continue to send pDelay requests on link down as well? I'm not sure that it makes sense?

pinealservo:
I'm not sure it matters whether they do; without link they'll stop getting network management messages and typically that means they'll go to sleep in fairly short order. They may want an explicit API at some point to manually stop for network management reasons, though--I'll have to look more closely.

I do know there's an explicit "AsCapable is always True" aspect to the automotive profile, though.

andrew-elder:
I was unaware of the events firing due to daemons attaching to the interface via libigb. We have a thread that monitors the link state and sends LINK_UP and LINK_DOWN events. Search for "watchNetLink". I'll create an issue to research "extra" link events being observed.

pinealservo:
Yeah, I traced through the netlink stuff at one point to see why it was happening; I think there's some level of hardware reset going on when a new userspace daemon attaches to the driver or something like that. This was especially causing issues for some people when something periodic was causing the events and we still had the link up/down reset call that leaked resources.

Just wanted to make sure you were aware that was happening with these changes here; switching AsCapable off and on momentarily could screw up things that don't have a bit of hysteresis in their state monitoring.

andrew-elder:
AsCapable always true for automotive change has been submitted.

pinealservo:
@kencarlino If you have a couple of minutes to look this over to make sure I didn't miss something relevant to the automotive profile, that'd help us get this merged quickly.

kencarlino:
These changes look good to me and correct the Automotive Profile wants asCapable always set to true.

pinealservo:
Thanks, @kencarlino!

PR #495, opened 2016-10-11
MRP: BUGFIX - memory leak in MSRP when checking for interesting streams

andrew-elder:
Note: This memory leak only occurs if TalkerAdv stream pruning is active
(cmdline option -p). Attributes that were selected to be pruned during
event processing were not being free'd.

### DIFF
--- a/daemons/mrpd/msrp.c
+++ b/daemons/mrpd/msrp.c
@@ -628,6 +628,9 @@ int msrp_event(int event, struct msrp_attribute *rattrib)
 #if LOG_MSRP
 			msrp_print_debug_info(event, attrib);
 #endif
+		} else {
+			/* free this attrib if we are not interested */
+			free(rattrib);
 		}
 
 		break;


daemons/mrpd/msrp.c
```diff
@@ -628,6 +628,9 @@ int msrp_event(int event, struct msrp_attribute *rattrib)
 #if LOG_MSRP
 			msrp_print_debug_info(event, attrib);
 #endif
+		} else {
+			/* free this attrib if we are not interested */
+			free(rattrib);
```
kencarlino: This is my first look at the mrpd code so this might be a common pattern and handled properly with an overloaded free() or something I am missing. The allocation pointed to by rattrib is freed but the pointer is never nulled out. Have concerns if it could be used later. However, I do see that in this case rattrib is passed into this function as a simple pointer so nulling it won't do any good outside of this function scope. The other concern is that on a quick search for this function msrp_event() that the caller of this function can also free this same allocation that was freed inside msrp_event() (at least in the one call case that I looked at). Concern would be a double free and potentially freeing some new allocation that might have occurred since the first free() or other general memory corruption.

andrew-elder: Thanks @kencarlino. You are correct about the potential double free(). While I was working on cpputest memory leak detection I generated exactly that "fault". I'm going to propose 3 steps. (1) accept this pull request, (2) I create a pull request with cpputest memory detection enabled that will detect multiple frees [I know this works because I have just had to address exactly that error - although in the test framework, not mrp source code], (3)  create a github issue to make code changes so that msrp_event() is always called with the same assumptions regarding responsibility for freeing rattrib.

Does that sound ok to you?

kencarlino: @andrew-elder Yes, this sounds like a good approach.

andrew-elder: See #498

PR #508, opened 2016-10-26
AVTP pipeline low latency changes.

MarcinMiklas:
AVTP pipeline can use igb launchtime feature (it will work only for one talker). 
To use that recompile pipeline using the command:

```
AVB_FEATURE_ENDPOINT=0 IGB_LAUNCHTIME_ENABLED=1 make all
```

New .ini file settings added allowing to use fixed timestamping mode, so no current gPTP wall time is used in AVTP packets, but timestamp which differs by packets interval from previous AVTP packet.

Also you can set thread affinity and priority in .ini file.

New ini config keys are:

```
fixed_timestamp
thread_rt_priority
thread_affinity
tx_blocking_in_intf
```

The last one is for allowing interface to pace packet sending, useful with live streams like ALSA talker.

This patch set also contains multichannel tonegen talker improvements.


andrew-elder:
I'm not sure that I know how to effectively review such a large set of changes. I'm inclined to accept the pull request in to open-avb-next and await on-going test results. Comments anyone?

pinealservo:
It may be easier to review if you pull the changes and look at the changes in an offline viewer, or look at them one at a time by clicking the 'Changes from all commits' text and selecting them individually or in small groups. It's unfortunate it won't show the changes inline in the "commits" view. The individual commits are pretty small and focused, it just took a lot of them to get it done.

MarcinMiklas:
There is a nice github trick to exclude whitespace changes from diffs. Just add `?w=1` to the and of the URL.

 So use this url for displaying file changes with whitespace changes ignored:

    https://github.com/AVnu/Open-AVB/pull/508/files?w=1

lib/avtp_pipeline/intf_echo/openavb_intf_echo.c
```diff
@@ -168,6 +171,10 @@ void openavbIntfEchoTxInitCB(media_q_t *pMediaQ)
 		}
 
 		pPvtData->Counter = 0;
+#if IGB_LAUNCHTIME_ENABLED
+		openavbAvtpTimeSetToWallTime(&pPvtData->walltime);
+		AVB_LOG_INFO(__func__);
```
kencarlino: Extremely minor comment. Why log a function name here?

lib/avtp_pipeline/intf_echo/openavb_intf_echo.c
```diff
@@ -209,7 +216,12 @@ bool openavbIntfEchoTxCB(media_q_t *pMediaQ)
 					printf("%s\n\r", (char *)pMediaQItem->pPubData); 
 			}
 
+#if IGB_LAUNCHTIME_ENABLED
+				openavbAvtpTimeAddUSec(&pPvtData->walltime, 125);
```
kencarlino: Adding a constant 125 usec means this is fixed for Class A (8000ps).  I believe this should be based on the tx rate specified in the .ini file for the stream.

kencarlino: Disregard. Fixed in a later commit.

kencarlino:
Looks good to me.

PR #510, opened 2016-10-27
Adding I2C_ENABLED make parameter for igb module

jfornal:
This is an urgent change request based on cooperation with a customer. The change introduces additional make parameter (I2C_ENABLED) which should be set to "1" when building a driver that will run with the use of I2C connection. This change prevents the driver from invoking set_sfp_media_type_82575() method each time. Previously even if I2C was disabled and not used, the driver invoked the method which returned an error and introduced an additional 300ms delay. Right now the delay is avoided.

The change has been tested on a customer-based HW setup and within an Intel product context

Some further refactoring of the e1000_get_media_type_82575() method might be needed in the future, but this change satisfies the basic need of reducing the time of igb_avb loading and saves over 300ms.


kmod/igb/e1000_82575.c
```diff
@@ -1797,7 +1799,13 @@ static s32 e1000_get_media_type_82575(struct e1000_hw *hw)
 		/* fall through for I2C based SGMII */
 	case E1000_CTRL_EXT_LINK_MODE_PCIE_SERDES:
 		/* read media type from SFP EEPROM */
+#ifdef I2C_ENABLED
+		printk(KERN_INFO "igb_avb I2C enabled - set_sfp_media_type_82575() called");
 		ret_val = e1000_set_sfp_media_type_82575(hw);
+#else
+		printk(KERN_INFO "igb_avb I2C disabled - set_sfp_media_type_82575() not necessary");
+		hw->phy.media_type = e1000_media_type_unknown;
```
jfornal: The ret_val is initialised with E1000_SUCCESS at the begining of e1000_get_media_type_82575() and it would be only changes in case of actually using set_sfp_media_type. So taking that into consideration there is no need to implicitly initialise the ret_val once again (so changing from E1000_SUCCESS to the same value) in #else statement - the method will still return SUCCESS so everything is consistent.

Don't you agree?

kencarlino: I agree, no problem. I viewed the larger scope of the function and see this SUCCESS being set.

PR #512, opened 2016-11-06
Fix gPTP link up/down detection. Check only the interface used by port

kencarlino:
This pull request fixes an issue that the link up and down detection in gPTP was not checking if the interface with the link event was the interface in use by gPTP. This would result in gPTP acting on incorrect link events on systems with multiple network interfaces. This issue was in Linux specific code.

Requesting expedited review of this pull request due to it blocking a customer release.


MarcinMiklas:
Looks good to me.

andrew-elder:
Looks good to me too. Merging it.

PR #516, opened 2016-11-14
Adding module parameter while performing insmod operation on igb_avb.ko

jfornal:
This is a change requested by one of Intel clients using the igb_avb driver. The default IGB_DEFAULT_TXD value used in terms of tx ring size was hardcoded to value of 256. It turned out during some tests that this is not enough. Therefore introducing a module parameter that can alter the tx ring size while performing insmod:

**sudo insmod igb_avb.ko tx_size=<value>**

Apart from changes in the igb_main.c itself, a similar module parameter has been added to the ethtool interface in case in is used, (tx_size_ethtool)

The default value stays the same, so performing **sudo insmod igb_avb.ko** without any additional parameters will introudce tx ring with size equal to 256 (as it was previosuly)


andrew-elder:
@jfornal - looks ok to me except for the accidental whitespace additions. Can you please double check for accidental tab additions?

jfornal:
I double-checked that and there are no additional whitespaces: in both igb_main.c and igb_ethtool.c there has been a deletion of the whitespace prior adding another one, so basically the number of whitespaces stays the same. The only place I can remove one whitespace is in igb_ethtool.c right after MODULE_PARM_DESC() - is that what you were referring to?

kmod/igb/igb_ethtool.c
```diff
@@ -1305,9 +1310,10 @@ static int igb_setup_desc_rings(struct igb_adapter *adapter)
 	struct igb_ring *rx_ring = &adapter->test_rx_ring;
 	struct e1000_hw *hw = &adapter->hw;
 	int ret_val;
-
+
```
andrew-elder: Extra whitespace here

kmod/igb/igb_main.c
```diff
@@ -2486,7 +2490,7 @@ static void igb_set_fw_version(struct igb_adapter *adapter)
 	struct e1000_fw_version fw;
 
 	e1000_get_fw_version(hw, &fw);
-
+
```
andrew-elder: Extra whitespace

kmod/igb/igb_main.c
```diff
@@ -4787,7 +4792,7 @@ static void igb_watchdog_task(struct work_struct *work)
 			default:
 				break;
 			}
-
+
```
andrew-elder: Extra whitespace

andrew-elder:
Thanks for the whitespace cleanup - all looks good to me now.

PR #522, opened 2016-11-19
fix avb_initialize_h1722_to_defaults bug

kyog02:
Fixed because initialization processing of sequence ID of frame 1722 could not be done.


andrew-elder:
This looks to me to be ok. I need someone else to review it/sign-off.

Issue #490, opened 2016-10-05
Remove remaining bare printfs from gPTP code


pinealservo:
This was brought to light during this review: https://github.com/AVnu/Open-AVB/commit/15c97ba4409d7f923dbc5ac8fd6c6a6e5349ba2e#commitcomment-19017533

We should replace those with the correct logging calls if they're necessary.

PR #526, opened 2016-11-23
Fix proposal to the issue 490

psow:
Solution proposal to issue #490 which is linked with printf calls within the gptp code.

Please note that this is supposed to serve as an introductory/first commit for a new member of Automotive society within Intel Technology Poland (Piotr Sowiński).


andrew-elder:
This looks to be ok to me. What do you think @bdthomsen ?

PR #551, opened 2017-02-02
gptp: fix wrong Follow_up Correction Field value in GPTP_LOG_VERBOSE

PawelModrzejewski:
correctionField doesn't need to be shifted again when logging as
it's done a few lines above.


andrew-elder:
I agree! Pull request looks good to me. Anyone else?

PR #560, opened 2017-02-27
Fix Travis build status URL in README

pinealservo:
It has been saying 'build unknown' because it's pointed at the AVnu/Open-AVB URL, which is no longer active on Travis. This points it at Avnu/OpenAvnu, which is active and shows the correct build status badge.

Also, changed one last "AVB" to "AVB/TSN".

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,8 +3,8 @@
 
 |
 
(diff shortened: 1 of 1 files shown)


andrew-elder:
Thanks @pinealservo !

PR #563, opened 2017-02-28
gPTP: redundant (false) LINKUP events filtering

PawelModrzejewski:
Mulitple LINKUP events are observed during interface manipulations
(e.g. adding or removing VLANs), so this change checks when
the ifi_flags gets IFF_RUNNING bit changed.

### DIFF
--- a/daemons/gptp/common/avbts_port.hpp
+++ b/daemons/gptp/common/avbts_port.hpp
@@ -237,6 +237,9 @@ typedef struct {
 	/* Set to true if the port is the grandmaster. Used for fixed GM in the the AVnu automotive profile */
 	bool testMode;
 
+	/* Set to true if the port's network interface is up. Used to filter false LINKUP/LINKDOWN events */
+	bool linkUp;
+
 	/* gPTP 10.2.4.4 */
 	char initialLogSyncInterval;
 
@@ -306,6 +309,7 @@ class IEEE1588Port {
 	unsigned sync_count;  // 0 for master, ++ for each sync receive as slave
 	// set to 0 when asCapable is false, increment for each pdelay recvd
 	unsigned pdelay_count;
+	bool linkUp;
 
 	/* Port Configuration */
 	unsigned char delay_mechanism;
@@ -1516,6 +1520,23 @@ class IEEE1588Port {
 	 */
 	bool getTestMode(void) {
 		return testMode;
+	}
+
+	/**
+	 * @brief  Sets the linkUp status
+	 * @param  bool of the linkUp status
+	 * @return void
+	 */
+	void setLinkUpState(bool state) {
+		linkUp = state;
+	}
+
+	/**
+	 * @brief  Gets the linkUp status
+	 * @return bool of the linkUp status
+	 */
+	bool getLinkUpState(void) {
+		return linkUp;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


daemons/gptp/linux/src/linux_hal_common.cpp
```diff
@@ -178,6 +178,7 @@ static void x_readEvent(int sockint, IEEE1588Port *pPort, int ifindex)
 	struct msghdr msg = { (void *) &snl, sizeof snl, &iov, 1, NULL, 0, 0 };
 	struct nlmsghdr *msgHdr;
 	struct ifinfomsg *ifi;
+	static unsigned last_ifi_flags;
```
andrew-elder: What is the initial state of last_ifi_flags?

daemons/gptp/linux/src/linux_hal_common.cpp
```diff
@@ -178,6 +178,7 @@ static void x_readEvent(int sockint, IEEE1588Port *pPort, int ifindex)
 	struct msghdr msg = { (void *) &snl, sizeof snl, &iov, 1, NULL, 0, 0 };
 	struct nlmsghdr *msgHdr;
 	struct ifinfomsg *ifi;
+	static unsigned last_ifi_flags = 0;
```
pinealservo: Doing it this way will break if multiple interfaces are ever used. We don't currently support multi-port usage, but we also don't want to make life difficult if we decide to implement it.

pinealservo:
Thanks for taking a look at this issue! Please let me know if you have any questions about the change request I asked for.

daemons/gptp/linux/src/linux_hal_common.cpp
PawelModrzejewski: Yes, it makes sense, thanks for comment.

andrew-elder:
I approved the change to init the flags field, but it really does need to be stored on a per port basis.

PawelModrzejewski:
linkUp per port added. Please review.

daemons/gptp/linux/src/linux_hal_common.cpp
```diff
@@ -205,11 +205,18 @@ static void x_readEvent(int sockint, IEEE1588Port *pPort, int ifindex)
 		if (msgHdr->nlmsg_type == RTM_NEWLINK) {
 			ifi = (struct ifinfomsg *)NLMSG_DATA(msgHdr);
 			if (ifi->ifi_index == ifindex) {
-				if ((ifi->ifi_flags & IFF_RUNNING)) {
-					pPort->processEvent(LINKUP);
+				bool newState = ifi->ifi_flags & IFF_RUNNING;
```
MarcinMiklas: Code will be easier to read if "newState" is renamed to "linkUp"

daemons/gptp/common/avbts_port.hpp
```diff
@@ -237,6 +237,9 @@ typedef struct {
 	/* Set to true if the port is the grandmaster. Used for fixed GM in the the AVnu automotive profile */
 	bool testMode;
 
+	/* Set to true if the IFF_RUNNING flag for port is up. Used to filter false LINKUP/LINKDOWN events */
```
andrew-elder: IFF_RUNNING is Linux specific, correct? This is common port code. Correct comment should be something like "Set to true is the port's network interface reports the link is "up"".

andrew-elder:
In general, I like the direction of these changes! Thanks for doing them.

PawelModrzejewski:
"newState" renamed to "linkUp" and comment in common part updated

andrew-elder:
I am now good with this pull request. Anyone else?

PR #565, opened 2017-03-08
S3 test and reboot test fixes - memory corruption/NULL ptr dereference

jfornal:
1. commit be07cb5
Fixing dma memory corruption during suspend - during S3 test there was a discovery of a memory corruption resulting in kernel panic. That was caused by memory beeing freed within suspend function. Keeping the memory in driver suspend prevents from the uexpected fatal behaviour

2. commit 7265634
Checking Tx and Rx queues number in igb_mapbuf - change due to kernel NULL pointer dereference in reboot test. There is a possibility that user space calls igb_ioctl_fila -> igb_mapbuf after the ring has been reset (and the tx/rx_rings have been NULL'ed). Therefore the patch check if tx and rx queues numbers are cleared to zero before accessing adapters tx_ring and rx_ring in igb_mapbuf

### DIFF
--- a/kmod/igb/igb_main.c
+++ b/kmod/igb/igb_main.c
@@ -10409,7 +10409,7 @@ static long igb_mapbuf(struct file *file, void __user *arg, int ring)
 		}
 		
 		if(!adapter->num_tx_queues) {
-			printk("tx ring freed\n");
+			printk("Tx_queues number cleared, tx ring freed - preventing from dereferencing NULL pointer\n");
 			return -EINVAL;
 		}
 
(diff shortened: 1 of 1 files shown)


kmod/igb/igb_main.c
```diff
@@ -10408,6 +10407,11 @@ static long igb_mapbuf(struct file *file, void __user *arg, int ring)
 			       req.queue);
 			return -EINVAL;
 		}
+		
+		if(!adapter->num_tx_queues) {
+			printk("tx ring freed\n");
```
andrew-elder: To my way of thinking the text in this message (and the rx one) should be more verbose.

andrew-elder:
Overall changes look good to me.

andrew-elder:
@jfornal  - I apologize, I wasn't explicit enough in my comment. My thinking was that there may be many sources of information being output to printk(), so the text output should identity the source of the message and perhaps the instance as well. I was thinking the source would be something like "igb" and the instance would be a device index, or a pointer - however IGB device indexes are maintained internally.

andrew-elder:
This all looks good to me. Need someone else to sign off.

PR #568, opened 2017-03-13
Automotive mode added to simple listener example

ghMichalW:
I would like to propose a change that allow to run simple listener in static configuration (no SRP). The change is dictated by a fact that Automotive AVB  that does not support stream protocol reservation.

### DIFF
--- a/examples/simple_listener/simple_listener.c
+++ b/examples/simple_listener/simple_listener.c
@@ -45,6 +45,7 @@ SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 			+ STREAM_ID_SIZE		\
 			+ SEVENTEEN22_HEADER_PART2_SIZE \
 			+ SIX1883_HEADER_SIZE)
+#define HEADER_SIZE_AUTOMOTIVE (38)
 #define SAMPLES_PER_SECOND (48000)
 #define SAMPLES_PER_FRAME (6)
 #define CHANNELS (2)
@@ -63,9 +64,17 @@ struct ethernet_header{
 static const char *version_str = "simple_listener v" VERSION_STR "\n"
     "Copyright (c) 2012, Intel Corporation\n";
 
+static const char *auto_dest_mac_prefix = "91:E0:F0:00:FE";
+
 pcap_t* glob_pcap_handle;
+
 u_char glob_ether_type[] = { 0x22, 0xf0 };
-SNDFILE* glob_snd_file;
+SNDFILE* glob_snd_file = NULL;
+u_char* glob_target_stream_id = NULL;
+
+u_char glob_automotive = 0;
+u_char glob_static_stream_idx = 1;
+u_char glob_static_stream_id[] = { 0x91, 0xE0, 0xF0, 0x00, 0xFE, 0x00, 0x00, 0x01 };
 
 static void help()
 {
@@ -75,17 +84,29 @@ static void help()
 		"Options:\n"
 		"    -h  show this message\n"
 		"    -i  specify interface for AVB connection\n"
+		"    -a  use automotive mode (no SRP)\n"
+		"          [Payload offset:%d, Freq:48kHz, Channels:2, SampPerFrame:6, AAF Int16 ]\n"
+		"          [MAC:%s:IDX, StreamId:0x%s0000IDX]\n"
+		"    -s  specify static stream index (IDX for automotive)\n"
 		"    -f  set the name of the output wav-file\n"
-		"\n" "%s" "\n", version_str);
+		"\n" "%s" "\n",
+		HEADER_SIZE_AUTOMOTIVE,
+		auto_dest_mac_prefix,
+		auto_dest_mac_prefix,
+		version_str);
 	exit(EXIT_FAILURE);
 }
 
 void pcap_callback(u_char* args, const struct pcap_pkthdr* packet_header, const u_char* packet)
 {
 	unsigned char* test_stream_id;
+	unsigned char* target_stream_id;
 	struct ethernet_header* eth_header;
-	uint32_t *buf;
+	// uint32_t *buf;
+	void *buf;
 	uint32_t frame[2] = { 0 , 0 };
+	uint16_t frame_16[2] = { 0 , 0 };
+
 	int i;
 	struct mrp_listener_ctx *ctx = (struct mrp_listener_ctx*) args;
 	(void) packet_header; /* unused */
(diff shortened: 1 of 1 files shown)


examples/simple_listener/simple_listener.c
```diff
@@ -165,26 +197,78 @@ void sigint_handler(int signum)
 #endif /* PCAP */
 
 #if LIBSND
-	sf_write_sync(glob_snd_file);
-	sf_close(glob_snd_file);
+  if (NULL != glob_snd_file)
```
pinealservo: The formatting in this chunk looks like it got mangled somehow.

examples/simple_listener/simple_listener.c
```diff
@@ -75,17 +84,29 @@ static void help()
...
 	exit(EXIT_FAILURE);
 }
 
 void pcap_callback(u_char* args, const struct pcap_pkthdr* packet_header, const u_char* packet)
 {
 	unsigned char* test_stream_id;
+	unsigned char* target_stream_id;
 	struct ethernet_header* eth_header;
-	uint32_t *buf;
+	// uint32_t *buf;
```
pinealservo: Please don't leave commented-out code.

examples/simple_listener/simple_listener.c
ghMichalW: Thank you for a great feedback. The reason why I propose that change is that my company is working on AVB Automotive stack and we have clients that are interested in using it with Open AVB. I have recently created  patch that allow to use Simple Listener for Automotive and I wanted to share it as maybe more people may be interested in it. However this avtp_pipeline listener may be a thing to go without all those changes (need to check).

Regarding your concerns:

Ad 1. Maybe you are right that I have made it too complex for example, I can limit the number of params to just 'a' for automotive and remove stream index

Ad 2. I agree that Automotive is a specific use-case but I believe that is worth to have it as an option for simple listener 

Ad 3. As much as I like to I can't currently work on development of matching simple_talker, but I believe that eventually I or someone else will update the talker with automotive option.

pinealservo: It's not so much the configuration complexity I'm concerned about, but the conditional blocks that are mixed in to the main program flow. These "simple talker/listener" demos are really only valuable as proof of concept and code examples, and the latter can be obscured as the code is extended to try to be more "useful".  For example, the simple_talker already suffers from trying to show both layer-2 and layer-3 streaming, with the logic for both stuck in-line inside conditionals. It really needs a clean-up to bring back its clarity as an example, but no one has time to do it!  I would like to avoid simple listener losing clarity in the first place.

One option for sharing this code without a lot of extra work on your part would be to simply make a new "simple_automotive_listener" as a separate example and host it in a fork of the OpenAvnu repository. Being separate from the existing examples, it would be easier to maintain, you wouldn't have to make any conditional logic, and you could make it as specific or general as you'd like. Your goals for the program and the amount of time you have to devote to it could be whatever you have in mind!

Another option, if you would really like to have it in the OpenAvnu repository, would be to provide the same basic functionality, but do so as separate "no SRP" and "use AAF" flags. Both of those options are potentially useful on their own to a wider group of people than just the automotive audience. Factoring the parsing code out of the main logic code, or perhaps providing two different packet handler functions for the different formats instead of interleaving them with conditionals, would help resolve my concerns about code complexity.

To sum up:

1. Additional options are fine, but try to make them fine-grained and generally useful if possible. A macro-option to set a different default set of options at once would be fine too as long as the configuration code doesn't become a mess in the process.
2. At least the default set of options for simple listener should match the default for simple talker; ideally both would get matching options at the same time, but that's not always feasible.
3. When adding the code to implement the options, try not to reduce the readability of the main program flow. This might mean breaking some code out into functions, or duplicating some parts of the code for different options, etc. I don't want the code that shows the basic aspects of writing an AVB listener to turn into a confusing mass of conditional logic! But I think this general goal can usually be accomplished without a whole lot of extra work.

I wrote a lot of text there, but I'm really not expecting massive changes to the patch you submitted in the pull request; I think some minor changes along the lines I suggested would make it okay with me. I appreciate the contribution, and I hope we can get it accepted without it becoming a major time sink for anyone.

ghMichalW:
I would like to have my code merged with OpenAvnu repository so I followed guideline and changed naming int no SRP and added AAF parameter. I have also created a separate callback for AAF packet handling to remove conditional code. Let me know what you think about it.

examples/simple_listener/simple_listener.c
```diff
@@ -264,19 +369,22 @@ int main(int argc, char *argv[])
...
-	sf_info->format = SF_FORMAT_WAV | SF_FORMAT_PCM_24;
+	sf_info->format = SF_FORMAT_WAV | sf_pcm_format;
 
 	if (0 == sf_format_check(sf_info))
 	{
 		fprintf(stderr, "Wrong format.");
 		return EXIT_FAILURE;
 	}
 
+  fprintf(stdout,"Output SampleRate:%d, Channels:%d, Format0x%X\n",
```
pinealservo: The indentation appears to be off here.

examples/simple_listener/simple_listener.c
```diff
@@ -293,8 +401,10 @@ int main(int argc, char *argv[])
 #if DEBUG
 	fprintf(stdout,"Got session pcap handler.\n");
 #endif /* DEBUG */
+
+  fprintf(stdout,"Create packet filter ether dst %s\n", dest_mac);
```
pinealservo: And also here.

ghMichalW:
I have submitted commit that fix the mentioned indentation

ghMichalW:
I would like to to ask to hold on with the merge for few days - I need to verify few things. Of course if there is anything else that should be fixed let me know, just wait with the merge to main. Thanks.

pinealservo:
We need someone else to take a look as well before we merge, but we'll also wait for you to give the OK.

andrew-elder:
I'm happy to review after @ghMichalW has given the go ahead.

ghMichalW:
Sorry for the delay but I am currently on a sick leave till the end of week. I will let you know next week.

andrew-elder:
Good catch @pinealservo.

ghMichalW:
Ops, my mistake. Of course I wanted to merge into open-avb-next branch. I have created new pull request targeting correct branch:
https://github.com/AVnu/OpenAvnu/pull/569

Unfortunately I am still on sick leave so final call for merge will be next week.

ghMichalW:
I have done the final checks and from my side the pull request is ready to merge:

https://github.com/AVnu/OpenAvnu/pull/569

PR #573, opened 2017-03-31
Common lib split

pinealservo:
This splits the `avb` library and its header file in the `lib/common` directory into several different component libraries so, for example, a program needing to interact with gPTP does not need to pull in code for SRP. The build systems for the examples and the avtp_pipeline are updated to take the split into account, and also to clean up and standardize the Makefiles a bit.

The code here is independent of the maap daemon code, so it can be reviewed separately from that pull request.

This is also the first step along a development line we've been pursuing that makes dependency on the IGB kernel module and support library optional where that's possible, especially in the avtp_pipeline code.

I'll be presenting the rest of that code in stages; both so that it can be more carefully reviewed and also because I'll need to disentangle it from other development we've done. We've been using it in a few different configurations across a couple of different platforms, though, so it should be in decent shape.


andrew-elder:
This pull request looks ok to me!

bdthomsen:
I approve it as well.

PR #575, opened 2017-04-06
Systemd watchdog

mgalka:
This is code adds systemd watchdog notifications to gPTP daemon.

The solution reads WatchdogSec parameter from service configuration file and notifies watchdog
every 0.5*WatchdogSec that gPTP daemon is alive.
If there is no watchdog configuration available or WatchdogSec == 0, watchdog notification won't run.

Watchdog configuration is read once during gPTP daemon startup.


bdthomsen:
Thanks for fixing the formatting, and for the contribution.  It looks good to me.

andrew-elder:
Agreed, merging

PR #583, opened 2017-04-20
Make dependency on igb library optional in avtp_pipeline

pinealservo:
This largely involves CMakeList.txt refactoring, but includes some cleanup of code so that things that depend on igb are explicit about it and things that depended on the previous common library now depend only on the parts of it they actually need.


andrew-elder:
All looks good to me. Anyone else?

bdthomsen:
It looks good to me as well.

PR #585, opened 2017-04-25
Separate common port functionality from Ethernet port functionality

christopher-s-hall:
802.11 support can be added with minimal code duplication


andrew-elder:
This seems to be ok to me. I have a few questions.
- it looks the EtherPort is a wired network port that would be a specific instance of CommonPort?
- many methods are now declared to be "const". Did this fix a compiler error of some sort?
- am I correct that an automatic code formatter was run on the changes?

christopher-s-hall:
-it looks the EtherPort is a wired network port that would be a specific instance of CommonPort?

Yes. [check here](https://github.com/christopher-s-hall/Open-AVB/blob/open-avb-next/daemons/gptp/common/ether_port.hpp#L88)

-many methods are now declared to be "const". Did this fix a compiler error of some sort?

Yes and no. I pulled some members from IEEE1588Port (e.g. _hw_timestamper) that I access in both CommonPort and EtherPort.  Making them protected avoids somewhat clunky accessor functions like getTimestamper() and leaves more of the original code intact. Making them const causes the compiler to flag re-assignment that would break things. An example, is that somewhere in the original port code _hw_timestamper was set to NULL. The compiler flagged this with the const keyword. At this point, it would be possible to remove the const designation, but I think using const is good practice in general.

-am I correct that an automatic code formatter was run on the changes?

I am the autoformatter. The entire patch only adds about 20 lines of executable code (this is almost all in CommonPort::processEvent()). I formatted common_port.cpp, but left ether_port.cpp mostly the same as ieee1588port.cpp. That is, ether_port.cpp is ieee1588port.cpp with a lot of stuff subtracted and minimal changes to get it to compile.

daemons/gptp/common/ieee1588.hpp
```diff
@@ -499,7 +501,8 @@ class HWTimestamper {
 	 * @param  frequency_offset Frequency offset
 	 * @return false
 	 */
-	virtual bool HWTimestamper_adjclockrate( float frequency_offset )
+	virtual bool HWTimestamper_adjclockrate
```
christopher-s-hall: Thanks for the feedback. I will check into those function descriptions. I tried to transpose all of those.

It may look funny, but const is fine for *adjclockrate because while it changes the state of the driver or device, it doesn't - currently at least - change the state of the timestamper object.

As I said below, I think using const is good practice.

christopher-s-hall:
I'll address @bdthomsen comment here for a larger audience. There actually was a redundant getClock() function that I removed. I added the description back. get/putTxLock() are virtual functions. I moved the description to the base class CommonPort. Does this make sense to document them once in the base class only?

christopher-s-hall:
I duplicated the license/copyright from the original ieee1588port.cpp. This seems like the right thing given the code has mostly been refactored, i.e. an "updated" license/copyright of some sort wouldn't be appropriate.

bdthomsen:
As long as what you are doing regarding the function descriptions is deliberate, and not an oversight, then I am good with it.

PR #588, opened 2017-05-02
Travis, switch to "trusty" linux distribution and remove gcc-4.6 support

andrew-elder:
Fix Travis build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+
+dist: trusty
+sudo: required
+
 language: c
 # git submodules workaround
 git:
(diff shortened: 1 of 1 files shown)


andrew-elder:
Well, that fix took much longer than I expected! I need someone else to sign-off on the change.

.travis.yml
```diff
@@ -9,22 +13,20 @@ before_install:
...
-env: BUILD_KERNEL=3.13.0-36-generic
+env: BUILD_KERNEL=4.4.0-75-generic
 install:
   - sudo apt-get update -qq
-  - sudo apt-get install -y libpcap-dev libpci-dev libsndfile1-dev libjack-dev linux-headers-3.13.0-36-generic cmake
+  - sudo apt-get install -y libpcap-dev libpci-dev libsndfile1-dev libjack-dev linux-headers-4.4.0-75-generic cmake
   - sudo apt-get install -y libgstreamer0.10-dev libgstreamer-plugins-base0.10-dev libasound2-dev
-  - sudo cp /usr/src/linux-headers-3.13.0-36/include/uapi/linux/ethtool.h /usr/include/linux
-  - sudo cp /usr/src/linux-headers-3.13.0-36/include/uapi/linux/ptp_clock.h /usr/include/linux
+  - sudo cp /usr/src/linux-headers-4.4.0-75/include/uapi/linux/ethtool.h /usr/include/linux
```
pinealservo: Are these still required? I thought they were there because the base system was so old that they weren't in the base includes, but they ought to be there in trusty. I guess we might need a newer version for some features, though.

pinealservo:
Thanks for updating this!

.travis.yml
andrew-elder: I'm not sure whether they are required. Do you think we should halt this this pull request to address that? In my opinion we merge this pull request as is and address other enhancements separately.

christopher-s-hall: We should get it running again and worry about details (cleanup, enhancements) later.

pinealservo: Yeah, I don't mean to halt the pull request, I was just curious if those were still needed. We can investigate later.

PR #590, opened 2017-05-05
Added YIELD command support to the MAAP daemon

bdthomsen:
This update provides an easy way to test the "yield" behavior of daemon clients.  This state occurs when a MAAP DEFEND PDU is received by the MAAP daemon from a device with a lower MAC Address that conflicts with a multicast address being defended by the daemon.  When the conflict occurs (either by receiving an applicable MAAP DEFEND PDU, or by using the "yield" command), the daemon client with the conflicting reservation will receive a MAAP_NOTIFY_YIELDED notification, followed by MAAP_NOTIFY_ACQUIRING and MAAP_NOTIFY_ACQUIRED notifications for the newly reserved MAAP multicast address.

### DIFF
--- a/daemons/maap/common/maap.c
+++ b/daemons/maap/common/maap.c
@@ -229,16 +229,16 @@ static int inform_status(Maap_Client *mc, const void *sender, int id, Range *ran
 	return 0;
 }
 
-static int inform_yielded(Maap_Client *mc, Range *range, int result) {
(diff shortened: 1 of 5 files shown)


andrew-elder:
This one looks ok to me too! Anyone else?

andrew-elder:
Thanks @bdthomsen!

PR #591, opened 2017-05-06
Windows 10 detection fix

bdthomsen:
Added a manifest indicating Windows 10 support to the Windows gPTP build, so that Windows 10 detection will work correctly.  Manifest support is available with CMake 3.4 and later.
Also changed ptp_message.cpp to address a x64 compile warning.
Note that these changes were only tested with Microsoft Visual Studio Premium 2012 Update 5, and not with any other compilers.


andrew-elder:
Looks ok to me. Merging....

PR #593, opened 2017-05-11
Added AVDECC information to gPTP daemon

bdthomsen:
Added information used by AVDECC to the gPTP daemon shared memory.
Also standardized references to IEEE 802.1AS-2011 in the comments.

### DIFF
--- a/daemons/gptp/windows/daemon_cl/windows_ipc.hpp
+++ b/daemons/gptp/windows/daemon_cl/windows_ipc.hpp
@@ -152,16 +152,39 @@ class WindowsNPipeMessage {
         NPIPE_MSG_TYPE getType() { return type; }
 };
 
+#ifndef PTP_CLOCK_IDENTITY_LENGTH
+#define PTP_CLOCK_IDENTITY_LENGTH 8		/*!< Size of a clock identifier stored in the ClockIndentity class, described at IEEE 802.1AS-2011 Clause 8.5.2.4*/
+#endif
+
 /**
  * @brief Provides an interface for the phase/frequency offsets
  */
 class Offset {
     public:
-        int64_t ml_phoffset;	/*!< Master to local phase offset */
-        FrequencyRatio ml_freqoffset;	/*!< Master to local frequency offset */
-        int64_t ls_phoffset;	/*!< Local to system phase offset */
-        FrequencyRatio ls_freqoffset;	/*!< Local to system frequency offset*/
-        uint64_t local_time;	/*!< Local time*/
+        int64_t ml_phoffset;					//!< Master to local phase offset
+        FrequencyRatio ml_freqoffset;			//!< Master to local frequency offset
+        int64_t ls_phoffset;					//!< Local to system phase offset
+        FrequencyRatio ls_freqoffset;			//!< Local to system frequency offset
+        uint64_t local_time;					//!< Local time
+
+        /* Current grandmaster information */
+        /* Referenced by the IEEE Std 1722.1-2013 AVDECC Discovery Protocol Data Unit (ADPDU) */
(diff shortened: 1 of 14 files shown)


christopher-s-hall:
This looks fine. Why do you #define the clock identity length? Is that because you're including these files in another application and want to make them more standalone.

daemons/gptp/windows/daemon_cl/windows_ipc.hpp
```diff
@@ -152,16 +152,39 @@ class WindowsNPipeMessage {
...
+        uint8_t gptp_domain_number;				//!< gPTP domain number
+
+        /* Grandmaster support for the network interface */
+        /* Referenced by the IEEE Std 1722.1-2013 AVDECC AVB_INTERFACE descriptor */
+        uint8_t  clock_identity[PTP_CLOCK_IDENTITY_LENGTH];	//!< The clock identity of the interface
+        uint8_t  priority1;						//!< The priority1 field of the grandmaster functionality of the interface, or 0xFF if not supported
+        uint8_t  clock_class;					//!< The clockClass field of the grandmaster functionality of the interface, or 0xFF if not supported
+        int16_t  offset_scaled_log_variance;	//!< The offsetScaledLogVariance field of the grandmaster functionality of the interface, or 0x0000 if not supported
+        uint8_t  clock_accuracy;				//!< The clockAccuracy field of the grandmaster functionality of the interface, or 0xFF if not supported
+        uint8_t  priority2;						//!< The priority2 field of the grandmaster functionality of the interface, or 0xFF if not supported
```
andrew-elder: May as well align all the dox comments vertically.

andrew-elder:
@bdthomsen - any reponse to @christopher-s-hall's question on why there is a #define for clock identity length?

bdthomsen:
I was trying to make the files consistent with ieee1588.hpp, and make it clear where the size "8" comes from.  I can remove that if there are are concerns.

andrew-elder:
I don't have a concern. Just making sure we close the loop. Are you ok @christopher-s-hall ?

christopher-s-hall:
My question just is why do we have the #define clock ID length in two places (ieee1588.hpp and ipcdef.hpp). Why not #include ieee1588.hpp in ipcdef.hpp.

It's not a big deal because the number will probably never change. If it did, the change would be required in two files.  I'm fine whichever way you think is right.

pinealservo:
@christopher-s-hall Most of the stuff in `ieee1588.hpp`, including transitive includes, wouldn't be appropriate for an external client.

As a compromise, `PTP_CLOCK_IDENTITY_LENGTH` could be moved to `ptptypes.hpp`, which is included by both `ieee1588.hpp` and `ipcdef.hpp`. I believe that would clean up the duplicate definition and avoid pulling in extra stuff for external users of the IPC header.

christopher-s-hall:
That makes sense. I just wanted to know the rationale. Either solution is fine - leave it as-is or move to ptptypes.hpp.

pinealservo:
Either way is fine with me, too.

andrew-elder:
@bdthomsen - the ball is in your court since it is your pull request. Do you want to leave the pull request as is, or move PTP_CLOCK_IDENTITY_LENGTH?

bdthomsen:
I went with the suggestion by @pinealservo.  Hopefully that addresses the comment by @christopher-s-hall.

andrew-elder:
This looks good to me. @pinealservo - do you want to merge it?

pinealservo:
I just noticed some odd behavior in our development branch vs. open-avb-next regarding timers; I want to verify that it's not caused by this patch before merging.

pinealservo:
OK, I'm pretty sure now it was a build hygiene issue of some sort. Should be good to go.

PR #594, opened 2017-05-11
Ignore twoStepFlag as per Table 11-4 of 802.1AS-2011

bdthomsen:
As the twoStepFlag value is not used, and should be ignored, added a change to not test that the flag is set.
This allows the gPTP daemon to work with switches, such as MOTU, that do not set this flag.

### DIFF
--- a/daemons/gptp/common/ptp_message.cpp
+++ b/daemons/gptp/common/ptp_message.cpp
@@ -840,7 +840,9 @@ void PTPMessageSync::processMessage( EtherPort *port )
 
 	port->incCounter_ieee8021AsPortStatRxSyncCount();
 
+#if CHECK_ASSIST_BIT
 	if( flags[PTP_ASSIST_BYTE] & (0x1<<PTP_ASSIST_BIT)) {
+#endif
 		PTPMessageSync *old_sync = port->getLastSync();
 
 		if (old_sync != NULL) {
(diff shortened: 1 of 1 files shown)


bdthomsen:
Does anyone have any comments regarding this pull request?

christopher-s-hall:
It looks OK. The default behavior is still to check the 'assist' bit?

andrew-elder:
Shouldn't this have been caught by UNH tests?

pinealservo:
No, the default behavior is now to ignore the PTP_ASSIST (1588v1 name) / twoStepFlag (1588v2 name) value. 802.1AS-2011 table 11-4 says "Reserved as TRUE, ignored on reception".

There doesn't appear to be any AVnu test mentioning table 11-4 or this particular bit of the flags field.

According to 1588v2, the twoStepFlag value indicates whether you've got your transmit timestamp embedded in your Sync/etc. or whether it's going to come in a FollowUp. The right 1588v2 response to a Sync with twoStepFlag set to False would be to extract the timestamp and process it, not to throw it away, so I can't see any justification for leaving the check on by default. It's not the correct behavior for 1588v2 or for 802.1AS.

pinealservo:
To be clear, this only manifested when using a not-AVnu-certified AVB Switch which obviously has a bug with respect to this flag. But it's probably the cheapest one available, and it seems silly to be incompatible with it by default when that behavior on our part is technically in violation of 802.1AS-2011.

christopher-s-hall:
I don't have any objections, but I think we could remove the check altogether. Either way is fine.

jdkoftinoff:
FYI if you upgrade your MOTU switch firmware this bug goes away.

andrew-elder:
I'm going to merge this pull request as is. There have been many comments, but none that block merging the pull request - at least in my opinion. If folks want to make further changes to the behavior, then another pull request can be submitted. Thanks to @bdthomsen for submitting this one.

PR #595, opened 2017-05-12
gPTP: add negative time jumps detection

PawelModrzejewski:
This change checks if time in follow_up messages is monotonic
and will discard “wrong” sync/follow_up to prevent calculating
incorrect values (ratio etc.). It was observed for some bridges
and for some cases, that such situation might happens (bridge issue)
and could lead to huge time instability, especially when syntonization
functionality is activated.

### DIFF
--- a/daemons/gptp/common/avbts_clock.hpp
+++ b/daemons/gptp/common/avbts_clock.hpp
@@ -61,6 +61,9 @@
    adjustment is performed */
 #define PHASE_ERROR_MAX_COUNT (6)
 
+/* Value returned by calcMasterLocalClockRateDifference() to indicate
+   detection of negative time jump in follow_up message */
+#define NEGATIVE_TIME_JUMP 0.0
 
 /**
  * @brief Provides the clock quality abstraction.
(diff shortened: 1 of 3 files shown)


mgalka:
Looks good to me.

christopher-s-hall:
Please make 0.0 a #define. I don't think it's very clear in the message processing code why 0.0 means the clock goes backward. Or maybe make the frequency a reference parameter and return an int (or bool or whatever) indicating some error occurred.

andrew-elder:
Seems to me when this occurs there is a major issue in the time aware system. So, I think LOG output should be LOG_ERROR instead of DEBUG as it is right now. Anything that is sending bad followup times can't really be relied on as a valid gPTP partner - at least to my way of thing.

PawelModrzejewski:
Thanks for the comments. I'll work on this to make it more clear. I would leave LOG_DEBUG inside calcMasterLocalClockRateDifference() and LOG_ERROR when the result is checked. I totally agree that using devices sending bad follow_up times is not the best approach - that change can help us sort out such cases :-)

pinealservo:
I think it would be more helpful to have the message with all the timestamp information in the log than just the message saying that a backward jump happened, which is all you can get if you log after checking the `0.0` result. I hope this will be a relatively rare occurrence, and having as much information as possible in the log might help to track down why it happened.

For that reason, I'd echo @andrew-elder in suggesting the LOG_DEBUG be changed to something that will be visible with normal log settings.

Thanks for the pull request, I'm sure this will help others as well.

PawelModrzejewski:
That was my second thought and it makes more sense, I'll do it like you suggested. Thanks!

andrew-elder:
Thanks for the update @PawelModrzejewski. I appreciate your efforts here!

PR #597, opened 2017-05-18
libigb: improve inter-process synchronization

tnishiox:
libigb used a posix named semaphore to protect concurrent accesses from
multiple processes. But since the posix semaphore cannot be automatically
released on process termination, if some process holding the semaphore
terminates without releasing it, other processes cannot acquire the semaphore
afterward. This could potentially cause a denial of service condition.
This patch replaces the posix named semaphore by a posix mutex.

In addition this patch adds a flag which prevents access to device after
calling igb_detach. Since igb_detach frees up associated resources, other
thread in multi-thread application should not access device once some thread
started resource clean-up by calling igb_detach. Otherwise invalid access to
freed memories could happen.

### DIFF
--- a/lib/igb/igb.c
+++ b/lib/igb/igb.c
@@ -1,6 +1,6 @@
 /******************************************************************************
 
-  Copyright (c) 2001-2016, Intel Corporation
+  Copyright (c) 2001-2017, Intel Corporation
   All rights reserved.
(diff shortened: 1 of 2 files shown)


andrew-elder:
Looks good to me too! Merging.

PR #600, opened 2017-05-19
gPTP: remove 'const' that breaks syntonization on linux

PawelModrzejewski:
With 'const' proper HWTimestamper_adjclockrate was not executed
and "Failed to adjust clock rate" error was flooding (with -S option).

### DIFF
--- a/daemons/gptp/common/ieee1588.hpp
+++ b/daemons/gptp/common/ieee1588.hpp
@@ -498,8 +498,7 @@ public:
 	 * @param  frequency_offset Frequency offset
 	 * @return false
 	 */
-	virtual bool HWTimestamper_adjclockrate
-	( float frequency_offset ) const
+	virtual bool HWTimestamper_adjclockrate( float frequency_offset )
 	{ return false; }
 
 	/**
--- a/daemons/gptp/common/ieee1588clock.cpp
+++ b/daemons/gptp/common/ieee1588clock.cpp
@@ -436,7 +436,7 @@ void IEEE1588Clock::setMasterOffset
 			if (port->getTestMode()) {
 				GPTP_LOG_STATUS("Adjust clock rate ppm:%f", _ppm);
 			}
-			if( !_timestamper->HWTimestamper_adjclockrate( _ppm )) {
+			if( !port->adjustClockRate( _ppm )) {
 				GPTP_LOG_ERROR( "Failed to adjust clock rate" );
 			}
 		}


pinealservo:
I couldn't figure out the mechanism for how this would break things. It was just recently introduced and was cited and looked at by at least one reviewer, and it was apparently justified at the time. So I looked around online C++ references and discovered why this happened.

The answer is apparently [const-overloading](http://www.dietmar-kuehl.de/mirror/c++-faq/const-correctness.html#faq-18.12), which seems like one of those features custom-designed to efficiently obliterate the feet of unsuspecting C++ programmers. Because the virtual function is a **const** member function but the linux-specific one of the same name is not a **const** member function, they are actually two different functions.

Since two of these functions exist, the one on the more specific class doesn't actually override the virtual one on the parent class; instead it's an entirely different member function that never gets called, so we always just get the 'false' return value.

While I believe **const**-correctness is a very good idea in a C++ project that has followed the discipline from the beginning, trying to add it in to an established C++ project appears to be an exercise in minefield-maneuvering. I'm not saying we shouldn't do it, but if we do, it should probably be done very carefully and reviewed closely for issues like this. I'm not experienced enough with the detailed gotchas of C++ to spot this sort of thing in a normal review.

christopher-s-hall:
Why not make LinuxTimestamperGeneric::HWTimestamper_adjclockrate() const?

andrew-elder:
Whow, that const overloading could catch anyone out on a bad day.

PawelModrzejewski:
To make LinuxTimestamperGeneric::HWTimestamper_adjclockrate() 'const', it is also needed to make LinuxTimestamperGeneric::Adjust() 'const'. I can push relevant changes if we agree to do it that way.

andrew-elder:
@PawelModrzejewski , yes please add the 'const' option as you outline above. Thanks for being flexible! I look forward to completely your pull request!

PawelModrzejewski:
To make LinuxTimestamperGeneric::HWTimestamper_adjclockrate() 'const', it is also needed to make LinuxTimestamperGeneric::Adjust() 'const'. I can push relevant changes if we agree to do it that way.

PawelModrzejewski:
I see some new issues after rebasing to the head of open-avb-next:
* common/ether_port.cpp:586:8: error: 'sync_timestamp' was not declared in this scope (fix ready)
* bool LinuxNetworkInterface::getLinkSpeed(int, uint32_t*): SIOCGIFNAME failed: Operation not supported (under investigation)
I will probably close this pull request and send new one.

PR #602, opened 2017-05-22
Feature avtp pipeline avdecc

bdthomsen:
Added basic AVDECC (IEEE 1722.1) support to the AVTP Pipeline implementation.


andrew-elder:
Thanks for this pull request!

andrew-elder:
I have one question. Why are there any LOCK/UNLOCK functions in any of the state machines? Surely the implementation should be organized so as to be single threaded?

bdthomsen:
The original implementation used a separate thread for each of the state machines, and my team kept with that model while fixing and enhancing it.  Going to a single-threaded model would require some rearchitecting that we're not currently planning to take on.

PR #605, opened 2017-05-24
gPTP: fix SIOCGIFNAME ioctl failure

tnishiox:
Use an INET socket rather than a NETLINK socket to get link speed
with ioctl() since NETLINK sockets do not support ioctl().

### DIFF
--- a/daemons/gptp/linux/src/linux_hal_common.cpp
+++ b/daemons/gptp/linux/src/linux_hal_common.cpp
@@ -310,6 +310,7 @@ void LinuxNetworkInterface::watchNetLink( CommonPort *iPort )
 {
 	fd_set netLinkFD;
 	int netLinkSocket;
+	int inetSocket;
 	struct sockaddr_nl addr;
 
 	EtherPort *pPort =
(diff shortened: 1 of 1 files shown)


tnishiox:
This patch will resolve the error message "bool LinuxNetworkInterface::getLinkSpeed(int, uint32_t*): SIOCGIFNAME failed: Operation not supported" which is stated in #600 at the bottom.

christopher-s-hall:
What kernel/driver are you using? I guess I'm not opposed to the merge, but I think we should understand why this is broken for some but not others.

tnishiox:
Hi,

I am using kernel v4.1.27 (IOW LTS kernel). If I see the kernel source code, the sock_no_ioctl() dummy handler is registered to the ops table of the netlink type sockets. 

https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/tree/net/netlink/af_netlink.c?h=v4.1.27#n3126
------------------------------------------------------------------------------------
static const struct proto_ops netlink_ops = {
	.family =	PF_NETLINK,
	....
	.ioctl =	sock_no_ioctl

The sock_no_ioctl() function simply returns -EOPNOTSUPP. Due to that daemon_cl came across the "Operation not supported" error when it tried the SIOCGIFNAME ioctl cmd.

https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/tree/net/core/sock.c?h=v4.1.27#n2141
------------------------------------------------------------------------------------
int sock_no_ioctl(struct socket *sock, unsigned int cmd, unsigned long arg)
{
	return -EOPNOTSUPP;
}

I realized that it may work with the latest mainline kernel v4.12 since now it has another ioctl handler named netlink_ioctl() which returns ENOIOCTLCMD. This will pass the ioctl cmd to the NIC drivers as fallback.

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/netlink/af_netlink.c?h=v4.12-rc2&id=08332893e37af6ae779367e78e444f8f9571511d#n1086
------------------------------------------------------------------------------------
static int netlink_ioctl(struct socket *sock, unsigned int cmd,
			 unsigned long arg)
{
	/* try to hand this ioctl down to the NIC drivers.
	 */
	return -ENOIOCTLCMD;
}
....
static const struct proto_ops netlink_ops = {
	.family =	PF_NETLINK,
	.ioctl =	netlink_ioctl,

andrew-elder:
@christopher-s-hall - any further comments?

PR #608, opened 2017-05-25
libigb: Fix some compiler warnings

davidcemin:
Also, removed some trailing spaces.

Signed-off-by: David Cemin <david.cemin@coveloz.com>


bdthomsen:
At first glance, the changes look fine.  However, there is a "You are viewing a condensed version of this merge commit" message when looking at the files changed, and a link from there leads to a whole boatload of old changes.  Has anyone seen this before?  Is this something we should be worried about?

davidcemin:
i did a merge from open-avb-next over mine, did my modifications and squashed the last two commits (to get rid of the merge over my branch message). 

I definitely didn't change anything else besides the changes on the igb.c file. (1 line plus removing some trailing spaces).

pinealservo:
I would suggest re-doing this. Reset your open-avb-next branch to the same state as the one here, then commit your change against that state and open a pull request again. I don't think this patch as-is would break anything, but it would put some weird stuff in the history.

You somehow made a merge commit instead of doing a fast-forward merge when bringing your open-avb-next up-to-date. Maybe this is because you checked in your change before realizing your open-avb-next was out-of-date?

A tactic I use to avoid weird git situations like this is to always do my local work on a short-lived task branch. Once I've tested it, I'll switch to whatever branch I want to make the pull request against, ensure it's up-to-date, and bring my task branch changes over. If my task branch was properly based on the one I'm going to push to github, it's an easy fast-forward merge. If I had some sort of mess-up, or if upstream changed significantly since I started the work, a fast-forward merge won't work.

If I can't fast-forward merge (i.e. `git merge --ff-only task-branch`) I will switch back to my task branch and rebase my task branch onto the current state of the long-lived branch it's based on. If there are a bunch of false-start commits or silly mistake cleanup commits I want to clean up, I do an interactive rebase instead to ensure that the commits make sense and have good commit messages. Github unfortunately reduces the utility of this, since it by default squashes the view of all commits to a single one in a pull request, but it's recommended practice for when you're committing to something like the Linux kernel.

After a rebase, you can switch back to the long-lived branch (e.g. open-avb-next) and the fast-forward merge should go through at that point; alternatively, you can push your task branch to your clone of the repo and do a pull request directly from your task branch to OpenAvnu/open-avb-next (or whichever long-lived branch you want to merge your changes into).

We haven't been requiring this in general, since it's a lot of hoop-jumping and frustrates people who aren't regular committers, but it makes the history log a lot easier to understand since it's not littered with merge commits, which introduce parallel timelines into the log.

davidcemin:
that's fair. I also think this is too messy and i dont mind redoing it.

davidcemin:
lets abandon this pull request. I just created a new one. Sorry for the noise, this was supposed to be simple :)

PR #620, opened 2017-06-05
gPTP: fix MSVC compile

andrew-elder:
Was doing some test compiles with MSVC and it didn't like the .set() operation on an unorderd-map, so I have rewritten it slightly.


christopher-s-hall:
set is not an unordered map operation, it's an operation on phy_delay_spec_t. In response to one of the comments in my last pull request, I changed the API for naming consistency to set_delay. I neglected to re-test this on the Windows build. I would prefer, changing to set_delay keeping consistency with the Linux code, but either should work.

andrew-elder:
Ok, I'll do it the linux way. I'm going to close this pull request.

PR #638, opened 2017-06-23
Open avb next

andrew-elder:
gPTP timer handling bugfix


andrew-elder:
I think I will go ahead and merge this. Any objections?

PR #640, opened 2017-06-27
mrpctl: read all pending messages

aditpape:
In each iteration mrpctl sends 3 messages but receives just one - this makes a lot of receive messages left unread in the socket and introduces a big delay in status updates.
This patch reads out all available messages during each cycle.

Signed-off-by: Andreas Pape <apape@de.adit-jv.com>


andrew-elder:
Thanks @aditpape. Can you please resubmit this against the open-avb-next branch as that is the approved path for bugfixes? I'm not going to insist, but if you know how to do a git rebase -i, it would be good to squash that syntax correction commit. Thanks again for this improvement.

PR #646, opened 2017-06-29
Rest of fix for gPTP intervals on ARM

pinealservo:
The previous fix correctly identified the problem but was incomplete--this
one has actually been verified to work. More `char` fields had to be changed
to `signed char` to avoid problems with negative numbers.

### DIFF
--- a/daemons/gptp/common/common_port.hpp
+++ b/daemons/gptp/common/common_port.hpp
@@ -314,9 +314,9 @@ private:
 	PortState port_state;
 	bool testMode;
 
-	char log_mean_sync_interval;
-	char log_mean_announce_interval;
-	char initialLogSyncInterval;
+	signed char log_mean_sync_interval;
+	signed char log_mean_announce_interval;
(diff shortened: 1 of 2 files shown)


pinealservo:
Sorry, my testing procedure for the last patch was flawed; I hadn't removed the command line parameters from the correct launch script, so it launched with command line parameters and worked correctly due to that instead of the incomplete fix.  I got the right script edited this time, so I'm pretty sure it's actually working to set the right default parameters now.

The Travis failure is unrelated to the code here; I'm not sure why it's doing that.

pinealservo:
Here's the Travis issue log for the apt repository causing the build fail: https://www.traviscistatus.com/incidents/z1317rvdkbsj

andrew-elder:
@pinealservo - thanks for the updated commit. @vitorroriz  - thanks for testing. Merging.

andrew-elder:
Looks like there is something wrong with travis at the moment, but I'm still going to merge this pull request.

PR #647, opened 2017-06-30
MAAP addresses reused, shaper daemon shutdown fix

bdthomsen:
Thread priority and affinity supported for Listeners
Added initial support to save useful information to the endpoint_save.ini file for re-use the next time the application is run.  By design, the ini file is only updated as needed.  Currently saving the assigned MAAP address so that the same address can be requested again.
If the shaper daemon receives a Ctrl-C signal, it shuts down as if the user called the quit ("-q") command, rather than just aborting.  This allows for cleanup of the current configuration to occur.
White space fixes.

### DIFF
--- a/daemons/gptp/common/common_port.cpp
+++ b/daemons/gptp/common/common_port.cpp
@@ -467,8 +467,18 @@ bool CommonPort::processStateChange( Event e )
 
 bool CommonPort::processSyncAnnounceTimeout( Event e )
 {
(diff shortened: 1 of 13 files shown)


bdthomsen:
This merge also moves the latest open-avb-next changes into the feature-avtp-pipeline-avdecc branch.  Apologies if the combined merge made it harder to follow.

PR #660, opened 2017-07-20
Reference: https://github.com/AVnu/OpenAvnu/issues/659

sachindes:
Error visible in case we start open_harness process again in same environment, is solved  by this fix.


andrew-elder:
This looks good to me. I need one more person to sign off.

sachindes:
Okay agree. As said issue was more prominent in case of exception or termination of openavb_harness. As a good practice, unlink should be used before bind.

We can skip this pull request.

PR #663, opened 2017-07-25
Feature avtp pipeline avdecc

bdthomsen:
If the Listener determines that the Talker it was connected to has stopped, it waits 5 seconds before attempting a fast connect to the Talker.  This is to deal with cases where the Talker would still advertise itself and/or respond to CONNECT_TX_COMMANDs while in the process of shutting down.
Stop the fast connect retry if the saved state information is deleted.
Also, fix to avoid divide-by-0 error seen occasionally when registering the stream at a bad time.


bdthomsen:
Just to be clear, this pull request is to pull updates from the feature-avtp-pipeline-avdecc branch into the open-avb-next branch, so it should go through the usual acceptance procedure.

bdthomsen:
Any comments on this pull request?

lib/avtp_pipeline/acmp/openavb_acmp_sm_listener.c
```diff
@@ -941,7 +942,18 @@ void openavbAcmpSMListenerSet_talkerTestFastConnect(
 		if (pDescriptor->stream &&
 				pDescriptor->fast_connect_status == OPENAVB_FAST_CONNECT_STATUS_TIMED_OUT &&
 				memcmp(pDescriptor->fast_connect_talker_entity_id, entity_id, 8) == 0) {
+			//
 			// We found a Talker matching the one we have been looking for.
+			//
+
+			// See if it is time to connect yet.
+			struct timespec currenttime;
```
andrew-elder: Isn't there an abstract time type that would be portable across platforms?

andrew-elder:
One small comment, but not major enough to prevent the merge.

lib/avtp_pipeline/acmp/openavb_acmp_sm_listener.c
bdthomsen: I didn't see one, and the code uses "struct timespec" in several other places outside the platform-specific directories, so the change would not be trivial.

andrew-elder: Fair enough then. I had to ask :) !

andrew-elder:
Can someone else please take 5 minutes to review this pull request. It's been open for a while now.

lib/avtp_pipeline/acmp/openavb_acmp_sm_listener.c
```diff
@@ -951,7 +963,8 @@ void openavbAcmpSMListenerSet_talkerTestFastConnect(
 				openavbAcmpSMListenerSet_doFastConnect(pDescriptor->stream, flags, talker_unique_id, talker_entity_id, controller_entity_id);
 			}
 			else {
-				AVB_LOGF_ERROR("Unexpected fast connect info for talker_entity_id=" ENTITYID_FORMAT, ENTITYID_ARGS(entity_id));
+				AVB_LOGF_DEBUG("Fast connect info for talker_entity_id=" ENTITYID_FORMAT " no longer valid", ENTITYID_ARGS(entity_id));
+				memset(pDescriptor->fast_connect_talker_entity_id, 0, 8);
```
kencarlino: I believe this is OK and that fast_connect_talker_entity_id is 8 bytes, however, consider using sizeof() here.

lib/avtp_pipeline/avdecc_msg/openavb_avdecc_msg_server.c
```diff
@@ -388,6 +388,9 @@ bool openavbAvdeccMsgSrvrHndlChangeNotificationFromClient(int avdeccMsgHandle, o
 										ENTITYID_ARGS(talker_entity_id),
 										ENTITYID_ARGS(controller_entity_id));
 								pDescriptor->fast_connect_status = OPENAVB_FAST_CONNECT_STATUS_TIMED_OUT;
+								memcpy(pDescriptor->fast_connect_talker_entity_id, talker_entity_id, 8);
```
kencarlino: Consider using sizeof here.

kencarlino:
No defects identified but added a couple of suggestions.

PR #673, opened 2017-08-22
Logging fixes

bdthomsen:
Cleaned up the general logging support for the maap and shaper daemons, and for AVTP Pipeline.
Changes to the rawsock ring logging when VERBOSE mode is enabled.


andrew-elder:
These changes look ok to me. Anyone else?

PR #678, opened 2017-08-23
SyncIntervalTimer lock changed to trylock

ghMichalW:
The change is required to prevent a deadlock that occurs
when PTPMessageSignalling is received in the middle of
processEvent SYNC_INTERVAL_TIMEOUT_EXPIRES

### DIFF
--- a/daemons/gptp/common/common_port.cpp
+++ b/daemons/gptp/common/common_port.cpp
@@ -311,7 +311,7 @@ void CommonPort::stopSyncReceiptTimer( void )
 void CommonPort::startSyncIntervalTimer
 ( long long unsigned int waitTime )
 {
-	syncIntervalTimerLock->lock();
+	if( syncIntervalTimerLock->trylock() == oslock_fail ) return;
 	clock->deleteEventTimerLocked(this, SYNC_INTERVAL_TIMEOUT_EXPIRES);
 	clock->addEventTimerLocked
 		(this, SYNC_INTERVAL_TIMEOUT_EXPIRES, waitTime);


andrew-elder:
Thanks. @christopher-s-hall - any comments before I merge?

ghMichalW:
@pinealservo Currently the syncIntervalTimerLock is only used in the code to start the timer so if trylock fails it means that some other thread is just starting the timer - so probably there is no point to re-start it again.

I was thinking about some more sophisticated solution but since the syncIntervalTimerLock is only for start timer I don't see a point in making it more complicated then it is necessary.

Anyway regarding the whole deadlock situation it is quite interesting and it took me some time to understand what is going on. The problem was hard to root cause because usually there are some sync messages already processed before deadlock occur and gPTP deamons report they are in sync but they just stop increment sync count due to one of the deamons entering deadlock. You can try to reproduce it by yourself with some heavy system load - I managed to reproduce it by starting gPTP deamon at system boot and after few reboots I usually get a deadlock. It is always in the situation when PTPMessageSignalling occurs during processEvent SYNC_INTERVAL_TIMEOUT_EXPIRES:

1. PTPMessageSignalling - acquire syncIntervalTimerLock lock (to re-start timer)
2. processEvent acquire TxLock and do some processing
3. PTPMessageSignalling - deleteEventTimerLocked  (to re-start timer)
4. PTPMessageSignalling - stop at acquire TimerQLock (in delete Timer)
5. processEvent - stop at acquire syncIntervalTimerLock
After that flow there is nothing more going on and the gPTP daemon has to be restarted

Maybe you can figure out some better solution for that issue - I am not an expert of timers or pthreads. In my case the trylock change solved the problem.

pinealservo:
@ghMichalW Well, I really appreciate that you put in the time to track that down!  Now that I've looked more closely, I think you're right and there isn't any additional danger to returning without note here. Although I don't really like it, it's not without a lot of precedent--in fact, addEventTimerLocked() does the very same thing in the same code path. Would probably take a significant amount of effort to get all the lock stuff cleaned up, and this solves a real bug, so I think it's worth merging.

ghMichalW:
@pinealservo Thank you. I am aware that my patch probably will not fix eventual problems with other timer event locks, but fortunately I didn't observe any other similar issues during stress tests.

I suspect there is some problem with Thread switching that block processEvent thread switch back to PTPSignallingMessage processing thread but I have no clue what may be the reason of that. If I will have more time I will try to take a closer look on that but for now to avoid sync issues I propose this patch.

andrew-elder:
Merging....

PR #683, opened 2017-08-24
PTP Port role restore on Link Up event

ghMichalW:
PTP Master/Slave port role restore added to
Link Up event processing to guarantee proper
gPTP daemon re-sync after Link loss situation
due to partner platform reboot. The PTP role
restore is especially important for the case
when platforms are connected directly with
ethernet cable - no switch in between.

### DIFF
--- a/daemons/gptp/common/ether_port.cpp
+++ b/daemons/gptp/common/ether_port.cpp
@@ -403,6 +403,14 @@ bool EtherPort::_processEvent( Event e )
 			GPTP_LOG_STATUS("LINKUP");
 		}
 
+		if( clock->getPriority1() == 255 || getPortState() == PTP_SLAVE ) {
+			becomeSlave( true );
+		} else if( getPortState() == PTP_MASTER ) {
+			becomeMaster( true );
+		} else {
+			startAnnounce();
+		}
+
 		if (automotive_profile) {
 			setAsCapable( true );
 


PawelModrzejewski:
Just wondering about calling startAnnounce() in automotive profile (but I haven't checked the code) ?

andrew-elder:
@pinealservo - should this pull request wait until you check out the automotive profile interaction?

pinealservo:
Here's my thinking after doing a quick check through the code:

We want to ensure that when running in automotive profile mode, we never start sending Announce messages. This code has a call to startAnnounce() that's not guarded with a check for automotive profile, so I wanted to ensure there's no chance of it being called when in automotive_profile mode.

The startAnnounce() call is only made if the previous two conditions aren't met; one of them will be met if getPortState() is PTP_MASTER or PTP_SLAVE.  The automotive profile code will, at initialization time, ensure that either PTP_MASTER or PTP_SLAVE is set--so if nothing ever changes the value from where it's initialized, there won't be a problem.

I think all the code paths that could change change the state are blocked off by automotive profile mode, but I haven't exhaustively verified it. I think it's probably safe though.

pinealservo:
In case it wasn't clear, I'm in favor of merging this now. I'll do it later today if there are no objections.

PR #688, opened 2017-09-04
gPTP Announce interval timer restart fixed

ghMichalW:
The change fix a problem of gPTP daemon synchronization
in non-automotive profile when daemon is  not ready at
start (asCapable=false) - for example due to link down.
In above situation with previous implementation the
announce interval timer will not restart on
ANNOUNCE_INTERVAL_TIMEOUT_EXPIRES event.


andrew-elder:
Anyone else have comments on this before I merge it?

PR #691, opened 2017-09-08
listen for announces on init and linkup

ethan-stewart-iol:
Currently, if starting two instances of daemon_cl, they do not connect to each other because neither listens for announce messages. This changes them from sending announces to listening. They then start announcing after an ANNOUNCE_RECEIPT_TIMEOUT_EXPIRES event.

### DIFF
--- a/daemons/gptp/common/common_port.cpp
+++ b/daemons/gptp/common/common_port.cpp
@@ -550,7 +550,7 @@ bool CommonPort::processEvent( Event e )
 
 		// If port has been configured as master or slave, run media
 		// specific configuration. If it hasn't been configured
-		// start announce message time
+		// start listening for announce messages
 		if( clock->getPriority1() == 255 ||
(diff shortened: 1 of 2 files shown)


andrew-elder:
I'm not sure that I understand how doing startAnnounce() immediately is any different from waiting a constant timer delay time and then starting to send announce messages. Shouldn't it end up being the same?

ethan-stewart-iol:
The issue is that it never started listening for announce messages. It would be ok to start the `ANNOUNCE_RECEIPT_TIMEOUT_EXPIRES` timer as well as the `startAnnounce()`, however when the `ANNOUNCE_RECEIPT_TIMEOUT_EXPIRES` event is triggered, `startAnnounce()` is called (which would be redundant, but should work fine).

andrew-elder:
Does anyone else have any comments on this pull request? I think is it ok.

PR #692, opened 2017-09-19
AAF 24 kHz support, cleanup

bdthomsen:
Added the missing 24 kHz rate to the list of supported rates.  (See  IEEE 1722-2016 Table 11)
Fix for case where audio payload is from two different media queue items (just in case).
Trace fix to use AVB_TRACE_MAP rather than AVB_TRACE_INTF.


andrew-elder:

andrew-elder:
This pull request requires 5 minutes of anyone's time to look over. Can someone please take a quick look and add their approval.

PR #693, opened 2017-09-19
Network interface string length consistency

bdthomsen:
More updates so that network interfaces include length for the rawsock type
prefix, and so the strncpy() commands use the extra length.
Also includes removal of some trailing whitespace.


andrew-elder:
This looks good to me. Anyone else?

PR #694, opened 2017-09-20
AAF returns appropriate max frame size for Talkers vs Listeners

bdthomsen:
Update to return more appropriate frame sizes based on Talker or Listener role.


lib/avtp_pipeline/map_aaf_audio/openavb_map_aaf_audio.c
```diff
@@ -433,6 +443,13 @@ void openavbMapAVTPAudioGenInitCB(media_q_t *pMediaQ)
 void openavbMapAVTPAudioTxInitCB(media_q_t *pMediaQ)
 {
 	AVB_TRACE_ENTRY(AVB_TRACE_MAP);
+	if (pMediaQ) {
+		media_q_pub_map_uncmp_audio_info_t *pPubMapInfo = pMediaQ->pPubMapInfo;
```
andrew-elder: pPubMapInfo does not seem to be used?

PR #695, opened 2017-09-25
Kernel update

jfornal:
This is a change that prevents compile error while trying to build the driver on kernel 4.9 version or higher. From that version, API of igb_ndo_set_vf_vlan method has changed - it requires additional argument. That code alteration maintains older kernel compatibility and has been created based on a similar change introduced to the regular igb driver.


andrew-elder:
Looks good to me too. Merging.

PR #698, opened 2017-09-29
Added logging support

bdthomsen:
AVTP Pipeline (openavb_host and openavb_harness) and AVDECC (openavb_avdecc)
have a new "-l" option to specify the filename of the log file to use.  If not
specified, then the behavior is unchanged, and stderr is used for logging.


andrew-elder:
Changes look good to me too - merging.

PR #700, opened 2017-10-04
Art and Logic apple vendor profile ptp changes.

rroussel:
These changes enable the support of the apple vendor ptp profile allowing interaction with apple proprietary ptp clocks. This implementation has been tested with the apple airplay SDK on a raspberry pi 3 running within a group of devices playing the same music stream.

To enable this feature one must build for the raspberry pi:
ARCH=RPI make clean all

NOTES:
1) The gptp_cfg.ini file must have the variables marked as "Apple Vendor PTP profile" set to the values within the default ini file for the code to work properly. Some tweaking of the priority1 value may be necessary to achieve the desired selection of best master clock for a given environment.

2) When in master state one must ensure that the unicast_send_nodes variable (in the ini file) is set with the IP addresses for the nodes to which ptp messages will be sent. This is a comma separated list of ipv4 or ipv6 addresses:
unicast_send_nodes = fe80::16f4:a1c4:8d8d:c5b1,192.168.0.188

3) An alternative to the use of the unicast_send_nodes ini variable is the use of the address registration message service. This allows other applications to register nodes at run time via socket communication. To use this interface set the ini variable address_registration_socket_port to the desired port number. For more detail about this interface please see addressapi.cpp

4) Though one can run the RPI build on any linux compatible hardware, the pulse per second interface will only run on the raspberry pi 3. This will set the physical pin number 11 (GPIO17) high once per second.


pinealservo:
Hi @rroussel,

This looks like an interesting patchset; if you are interested in getting the changes into the OpenAvnu repository, here are some recommendations to smooth things along:

1. Rebase your changes on top of open-avb-next
2. Break your changes into logical groups and submit as separate pull requests with the motivation for each

There seems to be quite a lot of refactoring to various files mixed in with feature-related changes; I am not opposed to cleanup or refactoring commits to make additional features more cleanly implementable, but it is difficult to tease out which changes are which when they're all interleaved together in a single pull request.

Due to how github manages pull requests, I'd recommend working on one small pull request at a time rather than trying to get the whole set queued up at once.

If you would rather just maintain your work as a separate fork, you are also welcome to make a pull request to a README that points to your repository for people interested in the Apple vendor PTP profile.

Thanks!

andrew-elder:
I would second the comments by @pinealservo. I would also thank you for the effort you have shown here - the changes look pretty good. I look forward to getting these changes merged in - just in smaller "bites"!

alecs2287:
Hello @rroussel,

Can I get the aPTP changes from somewhere? Or when do you plan to merge them in open-avb-next?

Thanks,
Alex

mxmora:
For the moment you can get the changes here:
https://github.com/rroussel/OpenAvnu/tree/ArtAndLogic-aPTP-changes

PR #701, opened 2017-10-06
appveyor

andrew-elder:


pinealservo:
Something is wrong with me, I am finding the watching of these streams of commits and the resulting build output strangely compelling.  Is this going to be the one that finally makes them both green?!

andrew-elder:
Sorry about all the noise. I wasn't set up to build linux gPTP here at home, so there was a bit much trial and error! But, on the bright side, both checks are now green!

PR #707, opened 2017-10-10
RFC: Simplify timeval & timespec redefinition fix

christopher-s-hall:
Eventually the additional #define can be removed, this is fixed in recent header files. I'm not sure if there are older header files that are even more broken that we want to support(?) I don't have very many installs to test against.


christopher-s-hall:
@pinealservo I think that sys/timex.h didn't exist at the time(?) I don't have any older Linux installs to consult. These headers have changed somewhat recently. Travis (and whichever distribution it's based on) requires the additional #define, but a Fedora 26 install doesn't.

PR #731, opened 2017-12-05
AVTP Pipeline Build Cleanup

bdthomsen:
Switched from defaulting from i210-specific to generic adapter support.
(Intel i210 will still work if the PCAP interface is used.)
GStreamer support defaults to being disabled, as the feature is not needed.
AVDECC binaries copied to the build/bin folder, so that .INI files can be shared more easily with the AVTP Pipeline binaries.
Explicitly sets CMAKE_TOOLCHAIN_FILE, which makes it easier to change the build type.
README file recommends using run_daemons.sh to start necessary daemons.


andrew-elder:
Looks good to me! Anyone else?

PR #732, opened 2017-12-05
Maintenance pull request from open-avb-next

andrew-elder:


andrew-elder:
I saw @bdthomsen approved, so I'm to merge as this is maintenance only.

PR #735, opened 2017-12-07
AVTP Pipeline Makefile cleanup

bdthomsen:
Removed avtp_avdecc_doc build references, as the documentation is the same as
that built for avtp_pipeline_doc.
Removed the unnecessary "lib" dependency for the AVTP Pipeline code.


bdthomsen:
The lib dependency can be problematic if building for a device where the IGB library does not compile, but is useful if someone enables IGB support when building the AVTP Pipeline module.  I'll see if I can find a better solution.

PR #748, opened 2018-01-12
Start PDelay calculation as fast as possible in Automotive Profile

PawelModrzejewski:
Trigger initial PDelay calculation without waiting full interval (def. 1 sec).
For Automotive Profile and fast initial Sync rate, we are losing that 1 sec.
due to discarded (not processed) Follow_up messages (delay = INVALID_LINKDELAY).


andrew-elder:
Thanks for reviewing guys. Merging...

PR #749, opened 2018-01-20
Introduce Libavtp

aguedes:
Hi all,

AVB applications/frameworks rely on AVTP protocol to transport audio, video, and control data from talkers to listeners. Due to the lack of a general purpose AVTP implementation, each application/framework implements its own version of a subset of AVTP so we end up having multiple and fragmented implementations of AVTP protocol that are not very reusable by new applications. That said, we created the libavtp project, an opensource, general purpose, standalone implementation of AVTP protocol, that we'd like to contribute to OpenAvnu.

The goal of libavtp is to provide a general purpose library that applications and frameworks can leverage to handle AVTP packetization. It defines packet structures, format field values and provides APIs to get/set packet fields. With libavtp, we converge all AVTP packetization-related efforts in one single project and benefit all other projects that use it.

Currently, libavtp supports only AVTP Audio Format (AAF). The infrastructure to extend it to support other formats such as Compressed Video Format (CVF) and Clock Reference Format (CRF) is already in place, and we plan to develop them in the open under OpenAvnu project.

General information and instructions on how to build libavtp are provided in the README.md file. Libavtp has unit tests for all public APIs, covering pretty much 100% of its source code. Instructions on how to run unit tests and generate coverage reports are provided in HACKING.md file.

Libavtp is under BSD-3-Clause License. For more information see LICENSE file.

Best regards,

Andre

### DIFF
--- a/lib/libavtp/include/avtp.h
+++ b/lib/libavtp/include/avtp.h
@@ -1,0 +1,97 @@
+/*
+ * Copyright (c) 2017, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *    * Redistributions of source code must retain the above copyright notice,
+ *      this list of conditions and the following disclaimer.
+ *    * Redistributions in binary form must reproduce the above copyright
+ *      notice, this list of conditions and the following disclaimer in the
+ *      documentation and/or other materials provided with the distribution.
+ *    * Neither the name of Intel Corporation nor the names of its contributors
+ *      may be used to endorse or promote products derived from this software
+ *      without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <errno.h>
+#include <stdint.h>
+
+/* AVTP subtypes values. For further information refer to section 4.4.3.2 from
+ * IEEE 1722-2016 spec.
+ */
+#define AVTP_SUBTYPE_61883_IIDC			0x00
+#define AVTP_SUBTYPE_MMA_STREAM			0x01
+#define AVTP_SUBTYPE_AAF			0x02
+#define AVTP_SUBTYPE_CVF			0x03
+#define AVTP_SUBTYPE_CRF			0x04
+#define AVTP_SUBTYPE_TSCF			0x05
+#define AVTP_SUBTYPE_SVF			0x06
+#define AVTP_SUBTYPE_RVF			0x07
+#define AVTP_SUBTYPE_AEF_CONTINUOUS		0x6E
+#define AVTP_SUBTYPE_VSF_STREAM			0x6F
+#define AVTP_SUBTYPE_EF_STREAM			0x7F
+#define AVTP_SUBTYPE_NTSCF			0x82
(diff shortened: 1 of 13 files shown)


andrew-elder:
Thank you for this pull request. I think it is great to have an easy method for accessing fields in an AVTP packet. My one concern is that there is extra memory access required to look up the bit field alignment and mask information before it is applied. On some of the processors I use this step is unacceptable. I would prefer a set of inline functions for accessing fields that would avoid the separate lookup stage. What do you think?

andrew-elder:
On another note, this pull request should be to open-avb-next, not master.

andrew-elder:
Also, please hook in to travis for automated build checking. See travis.yml in the top level dir.

aguedes:
Thanks for the feedback. I think we can get rid of the table and still keep the current API for accessing AVTP fields. For instance, in get_field_value(), instead of looking up the table, we could set 'mask' and 'shift' in each switch-case'.

I prefer the lookup table approach because it looks less error-prone to me. I see it brings an extra memory access but I wouldn't make compromises in code at this early stage if we are not sure it is indeed a bottleneck. Anyways, sure, we can change the approach if it is unacceptable to your platforms.

Just to make sure I understand the issue, could you please elaborate a bit on why it is unacceptable on those processors, and what processors are those? Are you running Linux on them?

Regarding hooking in travis, yes, I was planning to do that in a separate PR once this one was merged. Would that be fine?

Once the discussion is finished, I'll close this PR and open another one to open-avb-next branch. Thanks!

andrew-elder:
I run my AVTP implementation packet routines on a TI C674x DSP. It has a weird pipeline and loads take 6 cycles even if the data is in L1 cache. Even longer if it is not cached. I guess I don't see why you would go to the effort of creating simple C macros/functions that are not portable to multiple processor targets. It's just C after all. I certainly understand if the tester applications are not as portable - that is fine.

I don't see any advantage for the programmer (avtp lib user) between using
status = avtp_aaf_pdu_set(pdu, AVTP_AAF_FIELD_SV, 0x10);
vs
avtp_aaf_pdu_set_field_sv(pdu, 0x10);
I realize the later causes larger header files, but who would care? 

What do you think?

Great news on travis plans. Thanks!

Yes, lets continue on this PR until we get closure.

christopher-s-hall:
Using macros and string concatenation you should be able to preserve the first calling convention.

static int set_field_value(struct avtp_stream_pdu *pdu, enum avtp_aaf_field field, uint32_t val);

becomes:

static int _set_field_value(struct avtp_stream_pdu *pdu, uint32_t mask, uint8_t shift, uint32_t val);

#define AVTP_AAF_FIELD_SV_MASK (BITMASK(1) << SHIFT_SV)
#define AVTP_AAF_FIELD_SV_SHIFT (31 - 8)

#define SET_FIELD_VALUE( pdu, field, val ) \
  _set_field_value( pdu, field##MASK, field##SHIFT, val )

andrew-elder:
@christopher-s-hall - I like it. Plus the compiler will generate a compile time error if the xxxMASK is invalid, so it isn't any worse than using an enum.

aguedes:
Not sure if it was clear in my last reply but we can remove the extra memory access and still preserve the API, without macros or string concatenation. Maybe showing the code will make it clear. I'll rework the current code and post it here so you can take a look.

aguedes:
This new version should address your concern, @andrew-elder. Please let me know what you think so we can move forward with this PR.

andrew-elder:
@aguedes - why wouldn't you use string concatenation per the suggestion from @christopher-s-hall ? It is a much more optimal solution.

andrew-elder:
I need someone other than myself to approve this pull request before we merge.

lib/libavtp/include/avtp.h
```diff
@@ -0,0 +1,106 @@
...
+ */
+#define AVTP_SUBTYPE_61883_IIDC			0x00
+#define AVTP_SUBTYPE_MMA_STREAM			0x01
+#define AVTP_SUBTYPE_AAF			0x02
+#define AVTP_SUBTYPE_CVF			0x03
+#define AVTP_SUBTYPE_CRF			0x04
+#define AVTP_SUBTYPE_TSCF			0x05
+#define AVTP_SUBTYPE_SVF			0x06
+#define AVTP_SUBTYPE_RVF			0x07
+#define AVTP_SUBTYPE_AEF_CONTINUOUS		0x6E
```
andrew-elder: Alignment?

andrew-elder:
If someone else approves this, I'm not going to block it. It can be updated for efficiency at a later date if that becomes an issue for someone.

aguedes:
I don't think a macro-based API is a good approach for libavtp because we would have to expose several "internal" information on public headers that are not relevant to the user of the lib (e.g. MASK_*, SHIFT_*, macros from utils.h) so we'd end up "polluting" the public headers. Besides that, I feel like a macro-based API is trickier than a function-based API for lib users.

Sure we have room for optimizations, but IMO it is too early to jump into it. We should first have a straightforward API design and the best code maintainability we can. Later on, based on real use-cases numbers that show where the bottlenecks are, we should tackle optimizations and evaluate if the compromises in the code are worth it.

Hope this makes sense to you guys as well.

lib/libavtp/include/avtp.h
aguedes: Hmm, this looks aligned to me. Have you set 8-space tab?

andrew-elder: I'm just going by what I see in github

aguedes: I see. On vim the alignment is corret but github shows it misaligned for some reason :(

andrew-elder: I haven't read your coding guide. Can you just use spaces?

aguedes: Libavtp follows the coding style from the Linux kernel which uses tab for indentation.

If we take a took at the code using github UI (see https://github.com/aguedes/OpenAvnu/blob/libavtp/lib/libavtp/include/avtp.h) we can also verify the aligment is correct. In the diff it shows misaligned because of the '+' added to the beginning of the line.

aguedes:
OK, I'll close this PR and move to open-avb-next. Thanks!

PR #750, opened 2018-01-22
gptp: fix race condition during signal thread creation

PawelModrzejewski:
It was sometimes (2/100 + high cpu load) observed that signal thread
was not started (timerq->stop), which led to missing initial PDelay calculation and
thus to no Sync/Follow_up processing. Plus some debug logs added.


andrew-elder:
If @christopher-s-hall says it's ok, then it's good! I had a read and the changes look ok to me. Merging....

PR #751, opened 2018-01-27
Introduce libavtp

aguedes:
Moving the libavtp PR to open-avb-next branch as discussed in https://github.com/AVnu/OpenAvnu/pull/749.


andrew-elder:
Merging

PR #752, opened 2018-01-29
Checking whether the interface string is null-terminated

jfornal:
This a security issue (reported by one of Intel automotive customers) fix preventing from inappropriate printk() behavior in case of iface string passed as igb_bind() argument being not null-terminated.

Checking whether the last element of char req.iface is equal to '/0' fixes it. In case of an incorrect string, a regular warning is being generated and igb_bind() returns "invalid argument".


bdthomsen:
I'm not comfortable with this fix.  The string can be null-terminated even
if the GB_BIND_NAMESZ-1 character is non-zero, as any previous character
can be zero to indicate the end of a shorter string.  We should loop
through all the possibilities before returning an error indicating that the
string is not null-terminated.

andrew-elder:
What about using strnlen() ?

pinealservo:
It seems to me the right thing to do to resolve the security issue is to just set req.iface[IGB_BIND_NAMESZ-1] = 0 and not bother checking at all. If there's an earlier NULL, there's no problem. If there was a NULL exactly there, again no problem. If neither of those cases were true, there *was* a security problem, but there isn't anymore.

There may still be a user code problem, but that should result in a failed lookup, which will print out the full name field a few lines later, now NULL terminated and probably either truncated or containing some garbage.

christopher-s-hall:
Either @pinealservo or @andrew-elder solutions are fine. This code is not going to run very much. I would lean toward @andrew-elder's solution if this error case is expected to occur often. It avoids a potentially more expensive lookup for something that probably isn't there. Strnlen() probably also more effectively communicates the intent of the code.

andrew-elder:
@christopher-s-hall - funny, I think I prefer @pinealservo's suggestion over my one. I just didn't think of it!

christopher-s-hall:
@andrew-elder I don't have that much of an opinion either way. @pinealservo's suggestion is fine with the appropriate comment.

andrew-elder:
Merging

PR #755, opened 2018-02-02
travis: Hook in libavtp

aguedes:
This patch hooks libavtp into travis in order to have the CI system
building libavtp and example applications, as well as running the unit
tests.

Libavtp dependencies are installed manually because the packages
provided by Ubuntu Trusty are very old versions and incompatible. Also,
likewise ethtool.h and ptp_clock.h, this patch copies the if_ether.h
header from linux-headers-4.4.0-75 directory to /usr/include since the
version provided by Trusty is very old and doesn't define the ETH_P_TSN
macro.

Finally, this patch sets CFLAGS=-Wno-missing-braces when building
libavtp in order to silent a warning generated by gcc. Travis is using a
very old gcc version (4.8) which wrongly warns about universal zero
initializer (more details about this on
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53119).


andrew-elder:
Thanks!

PR #760, opened 2018-02-26
libavtp: Add pkg-config support

aguedes:
This patch adds support for pkg-config in order to make it easier to
other projects to link against libavtp.

We leverage Meson builtin support for generating and installing the
.pc file.


andrew-elder:
These changes look good to me. Anyone else?

PR #761, opened 2018-02-27
Kernel update support

jfornal:
Adding igb_avb support for kernel 4.13 (previously it caused compile errors):
a)  replacing pci_enable_msi as it's been officially removed from kernel sources
b) changing invocation of igb_vm_fault - API has changed and there is no need to pass vm_area_struct any more
c) including miscdevice.h in order to ensure that the compiler recognizes the igb_miscdev structure elements


jfornal:
Notice additional things that require adding - closing current pull request.

PR #763, opened 2018-03-05
Resolving Klocwork Static Code Analysis issues

jfornal:
Recently Klocwork tool has been applied to the igb_avb driver code and as a consequence I'm pushing some resolved issues reported by that tool:

a) e1000_82575.c : e1000_clock_in_i2c_byte always returns "0" and other methods from the i2c API called within this method do the same. Therefore there is no sense in checking whether the returned value is !0 as it will never happen

b) igb_ethtool.c : 
     1. changing phy.type into phy.id as it was compared with the wrong define value
     2. igb_setup_loopback_test returns always 0 so there is no point I checking the output. The if   statement applied to igb_run loopback_test instead (more meaningful)
     3. Retrieve "*" operator not needed in case of memcpy and passing table igb_gstrings_test address

c) igb_main.c : igb_up returns always 0, so if(igb_up) will never occur

d) igb_param.c : wrong logic inside if statement. OR statement will not prevent from entering the code inside "if" in a proper way. AND is the desired one

### DIFF
--- a/kmod/igb/e1000_82575.c
+++ b/kmod/igb/e1000_82575.c
@@ -3218,8 +3218,6 @@ s32 e1000_read_i2c_byte_generic(struct e1000_hw *hw, u8 byte_offset,
 			goto fail;
 
 		status = e1000_clock_in_i2c_byte(hw, data);
-		if (status != E1000_SUCCESS)
-			goto fail;
 
 		status = e1000_clock_out_i2c_bit(hw, nack);
 		if (status != E1000_SUCCESS)
--- a/kmod/igb/e1000_82575.c
+++ b/kmod/igb/e1000_82575.c
@@ -98,10 +98,10 @@ static void e1000_clear_vfta_i350(struct e1000_hw *hw);
 
 static void e1000_i2c_start(struct e1000_hw *hw);
 static void e1000_i2c_stop(struct e1000_hw *hw);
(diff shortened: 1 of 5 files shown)


christopher-s-hall:
Would it be better to change the function return type to void for functions that return a 'constant' int?

andrew-elder:
@jfornal - any response to @christopher-s-hall s comment?

jfornal:
Yes, of course - I agree with the suggestions. I implemented them, just waiting for the Protex Legal scan to be completed and then I'll submit the altered code.

andrew-elder:
I would hesitate to remove error code returns for I2C transactions. Are we sure the I2C operations in question can never return errors? Here I'm referring to the I2C protocol itself, not the way it is currently implemented.

jfornal:
I checked the code and in case of "in" I2C transactions there is no error/status reporting, so it is safe from the igb_avb perspective. The thing that I removed was "return 0" in each case - without any logic behind that. So given the fact that it is an "in" transaction method and the second argument I would say that it there is no reason not to change it into a void method.

pinealservo:
Wow, bit-banging i2c code.  There aren't typically any error conditions from a lower level; it's just toggling and reading GPIO pins. You *could* conceivably do more checking for valid protocol states, but I don't think you'd be able to detect anything within those functions, but rather by checking whether they had the desired effect or not.

The functions are also all static, and will probably end up inlined anyway. I don't have a problem with the changes.

PR #764, opened 2018-03-07
Port IGB_AVB to Linux v4.16-rc4

milloni:
This PR makes changes necessary to build the IGB_AVB kernel driver with Linux v4.16-rc4 (and hopefully 4.16, when it is released).

Changes:
- `ACCESS_ONCE` is no longer available, replaced with `READ_ONCE`
- `setup_timer` is no longer available, replaced with `timer_setup`


andrew-elder:
Thanks. This looks ok to me. Not sure why tavis is broken, but I think that is separate issue.

jfornal:
I approve as well - lately experienced similar issues while compiling under kernels newer than 4.15. This saves some of the work I would have to do anyway : )

PR #767, opened 2018-03-15
travis: Stick to meson version 0.44

aguedes:
Meson 0.45 was released recently and it requires Python 3.5 which is not
installed in travis environment. This patch changes .travis.yml so
version 0.44 is installed instead.


andrew-elder:
Great! I'm going to go ahead and merge this as the change is trivial.

PR #777, opened 2018-04-18
gptp : remove unnecessary sequence that try to delete pdelay request object.

jongwook-lee:
A object of pdelay_req is managed(create&delete) at ethe_port.
but it also could be deleted at a step for processing of pdelay_resp.

Those sequence should be deleted for some reasons below:
1. If pdelay_req is trigged in the middle of processing pdelay_resp, then 'null point exception' could be happened.
2. The available pdelay_req would(should) be kept until sending next pdelay_req. Becuase it is possible to receive correct pdelay_resp even after unexpected error at processing pdelay_rest.


andrew-elder:
Please resubmit pull request against branch open-avb-next. Thanks,

Issue #776, opened 2018-04-16
Possible dead lock when handling SYNC_RECEIPT_TIMEOUT_EXPIRES event


freddeng:
Hi,

I am using the PTP daemon (from https://github.com/rroussel/OpenAvnu) with AirPlay2 SDK and found a possible dead lock when SYNC_RECEIPT_TIMEOUT_EXPIRES is received.

Although this was found in the branch (https://github.com/rroussel/OpenAvnu), I assume the Master branch will be the same as this part of the code is the same.

What happens is that 
1. CommonPort::stopSyncReceiptTimer() or CommonPort::startSyncReceiptTimer() gets hold of syncReceiptTimerLock before calling clock->deleteEventTimerLocked()/addEventTimerLocked();
2. then SYNC_RECEIPT_TIMEOUT_EXPIRES is received and handled in LinuxTimerQueueHandler() and that gets hold of timerq->lock, later in the callback CommonPort::processSyncAnnounceTimeout() it will try to call startSyncReceiptTimer() which be blocked by step 1, in the meantime step 1 will be also blocked by the timer callback because it waits for timerq->lock.

I think normally startSyncReceiptTimer/stopSyncReceiptTimer will finish in a very short time so maybe it's not very obvious. But in my environment it looks like startSyncReceiptTimer/stopSyncReceiptTimer sometimes take ~100 ms to execute "syncReceiptTimerLock->lock()"so this happens quite often.

In my opinion, we could fix this by using just one lock, the one in TimerQueue, like add a new method in IEEE1588Clock called "readdEventTimerLocked":
void IEEE1588Clock::readdEventTimerLocked
(CommonPort *target, Event e, unsigned long long time_ns)
{
    if( getTimerQLock() == oslock_fail ) return;
	deleteEventTimer( target, e);
	addEventTimer( target, e, time_ns );
    if( putTimerQLock() == oslock_fail ) return;
}
And remove syncReceiptTimerLock completely from CommonPort to avoid this dead lock but I do not know if that is going to cause other side effects. 
Also maybe similar dead lock could happen for CommonPort::startSyncIntervalTimer and CommonPort::startAnnounceIntervalTimer but I am not entirely sure.

christopher-s-hall:
I'll have to take a closer look at the code, but I suspect syncReceiptTimerLock and friends were added when we added processing for signalling messages that modify the timers outside of the handler. I'm not immediately sure of the implications of removing those locks.

Would something like the following work:

[...]

PR #779, opened 2018-04-18
Fix the dead lock when handling SYNC_RECEIPT_TIMEOUT_EXPIRES event

freddeng:
This is to fix issue #776 by adding timer queue lock in startSyncReceiptTimer and stopSyncReceiptTimer.

### DIFF
--- a/daemons/gptp/common/avbts_clock.hpp
+++ b/daemons/gptp/common/avbts_clock.hpp
@@ -162,7 +162,9 @@ private:
 
     OSLock *timerq_lock;
 
-	/**
+public:
+	
+    /**
 	 * @brief  Add a new event to the timer queue
 	 * @param  target EtherPort target
 	 * @param  e Event to be added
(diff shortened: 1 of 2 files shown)


christopher-s-hall:
Can you use addEventTimer() and deleteEventTimer(), rather than the *Locked() version? The lock is already held before calling the lock function. I see the lock is recursive, but I think that's a mistake - probably a workaround for some signal handler that doesn't exist.

andrew-elder:
@christopher-s-hall - thanks for the comments!

freddeng:
Hi @christopher-s-hall , as far as I can see addEventTimer and deleteEventTimer are private so I didn't use them.

christopher-s-hall:
I'd say make them public.

freddeng:
OK, I will give that a go. How do I proceed on this one, should I close it and create a new pull request to include all changes?

andrew-elder:
You can add changes on top of this patch if you like.

andrew-elder:
This change now looks correct to me. What do you think @christopher-s-hall ?

PR #780, opened 2018-04-18
gptp : remove unnecessary sequence that try to delete pdelay request object.

jongwook-lee:
A object of pdelay_req is managed(create&delete) at ethe_port.
but it also could be deleted at a step for processing of pdelay_resp.
Thus, there is a race condition.

Those sequence should be deleted for some reasons below:
1. If pdelay_req is trigged in the middle of processing pdelay_resp, then 'null point exception' could be happened.
2. The available pdelay_req would(should) be kept until sending next pdelay_req. Becuase it is possible to receive correct pdelay_resp even after unexpected error at processing pdelay_rest.

### DIFF
--- a/daemons/gptp/common/ptp_message.cpp
+++ b/daemons/gptp/common/ptp_message.cpp
@@ -1818,8 +1818,6 @@ void PTPMessagePathDelayRespFollowUp::processMessage
 	port->setPeerOffset( request_tx_timestamp, remote_req_rx_timestamp );
 
  abort:
-	delete req;
-	eport->setLastPDelayReq(NULL);
 	delete resp;
 	eport->setLastPDelayResp(NULL);
 


andrew-elder:
@christopher-s-hall - any comments? It looks ok to me.

bdthomsen:
The "goto abort" on line 1573 is problematic.  eport is NULL, and the req and resp pointers are undefined, so nothing good will come from going to the abort logic from that point.  I suspect a simple "return" is the correct behavior.

jongwook-lee:
thank you @bdthomsen for feedback.
The exception could be happened when normal status, you can see the details at first comment i made.
Actually, It is caused by that two threads manage one resource(the instance of req)..
So we need  to choose one of two options to resolve it:
1. add a algorithm for shared memory such as mutex, semaphore...
2. just delete it from one thread if it is unnecessary.
and i've just suggested second option.

andrew-elder:
While the comment by @bdthomsen is a good one, how about we create a separate issue for it? It seems like an additional observation that is outside the scope of this pull request.

andrew-elder:
I just reread the comments - my previous comment may be incorrect!

bdthomsen:
I'm fine with handling it either way.

christopher-s-hall:
I'm happy with it as-is

PR #792, opened 2018-06-05
Remove verbiage from OpenAVnu Readme referring to Intel submission checking

christopher-s-hall:
Intel does not check any non-Intel code submissions

### DIFF
--- a/daemons/gptp/common/common_tstamper.hpp
+++ b/daemons/gptp/common/common_tstamper.hpp
@@ -73,6 +73,15 @@ public:
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


andrew-elder:
This pull request has 2 distinct commits. @christopher-s-hall, was the intent to update readme AND add the monotonic raw clock option? Do you mind splitting in to 2 pull requests? If you like you can change the readme, do the pull request and merge it yourself. It's only a text change and it looked fine to me. Then we can leave the monotonic clock part for a more detailed review.

PR #795, opened 2018-06-11
gptp: Add monotonic raw clock option

christopher-s-hall:
Timestamp clocks that implement the PTP_SYS_OFFSET_PRECISE ioctl return
 cross-timestamps with respect to monotonic raw and realtime
Add configuration file option allowing clock selection for system/local
 clock relation

The default clock type is Realtime which is equivalent to the
 CLOCK_REALTIME option to clock_gettime
Optionally, MonotonicRaw (equivalent to CLOCK_MONOTONIC_RAW) can be
 specified in the gPTP daemon configuration file:

[clock]

SystemClock = MonotonicRaw

Fixed Makefile issue where some shell configurations use built-in 'echo'
 and others don't
Forces all to use /bin/echo

### DIFF
--- a/daemons/gptp/common/gptp_cfg.hpp
+++ b/daemons/gptp/common/gptp_cfg.hpp
@@ -66,23 +66,23 @@ class GptpIniParser
          */
         typedef struct
         {
-            /*ptp data set*/
-            unsigned char priority1;
+		/*ptp data set*/
+		unsigned char priority1;
 
 		/* Clock data set */
 		char systemClockDesc[MAX_CLOCK_DESC_LEN+1];
 
-            /*port data set*/
-            unsigned int announceReceiptTimeout;
-            unsigned int syncReceiptTimeout;
-            unsigned int syncReceiptThresh;		//!< Number of wrong sync messages that will trigger a switch to master
-            int64_t neighborPropDelayThresh;
-            unsigned int seqIdAsCapableThresh;
-            uint16_t lostPdelayRespThresh;
-            PortState port_state;
+		/*port data set*/
+		unsigned int announceReceiptTimeout;
+		unsigned int syncReceiptTimeout;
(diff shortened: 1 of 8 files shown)


daemons/gptp/gptp_cfg.ini
```diff
@@ -6,6 +6,11 @@
 # The lower the number, the higher the priority for the BMCA.
 priority1 = 248
 
+# Option to specify system clock type
```
andrew-elder: Can we add some more details here. Specifically, we should list the currently supported options, the default option and what the options mean. The text that is part of the pull request has great information in it, but I think that 6 months from now readers will prefer to read an updated comment here.

andrew-elder:
Looks good to me. One small comment update requested. Can anyone else review this?

daemons/gptp/common/gptp_cfg.hpp
```diff
@@ -68,6 +69,9 @@ class GptpIniParser
             /*ptp data set*/
             unsigned char priority1;
 
+		/* Clock data set */
```
pinealservo: It seems like there might be a tabs vs. spaces conflict in this file; I'm not sure whether it'd be better to just make the change match or if the file itself has a mismatched convention.

daemons/gptp/gptp_cfg.ini
pinealservo: Yes, I found comments in the example ini files very helpful when trying to actually use the avtp_pipeline code, so it'd be a nice help for users.

christopher-s-hall:
I added more comments to the config file and I added a second commit to fix the mixed whitespace in gptp_cfg.hpp at least for that struct. My spacing was/is correct (tabs only). What was there wasn't.

andrew-elder:
@pinealservo - looks good to me. If you agree, please go ahead and merge. Thanks for the updates @christopher-s-hall

PR #827, opened 2018-08-31
gptp: Fix error logging for event pkt without timestamp

tgorochowik:
This actually changes two things:
  * Use \t instead of \n similarly to what other error log messages do
  * Use the correct msg variable (err_msg instead of msg which is used
    for something else)


tgorochowik:
Just noticed the code has been moved to a separate repo.

I will send the same patch over there.

PR #836, opened 2018-10-01
Open avb next

andrew-elder:
Split off gptp, igb and libavtp into their own repos.


andrew-elder:
Any comments on this pull request? I'll complete it on Friday if I don't see any comments.

PR #841, opened 2018-10-12
Add h264_timestamp size into the H264 data length, remove it from the…

yesbman:
Add h264_timestamp size into the H264 data length, remove it from the AVTP header (IEEE 1722-2016 8.5.3.1 h264_timestamp).
The stream data length of AVTP header stands for the size of h264 timestamp + h264 payload.

- Issue : https://github.com/AVnu/OpenAvnu/issues/832

1. Author : Byoungman An
2. Title : Open-AVB


yesbman:
I have changed the base branch from master to open-avb-next as you requested.

PR #842, opened 2018-10-15
Add h264_timestamp size into the H264 data length

yesbman:
As andrew-elder has been mentioned from https://github.com/AVnu/OpenAvnu/pull/841#issuecomment-429324944, I pull request at the open-avb-next branch.

Add h264_timestamp size into the H264 data length, remove it from the AVTP header (IEEE 1722-2016 8.5.3.1 h264_timestamp).
The stream data length of AVTP header stands for the size of h264 timestamp + h264 payload.


yesbman:
This pull request is not necessary because it is rebased at https://github.com/AVnu/OpenAvnu/pull/841; thus, I have just closed it.

PR #888, opened 2019-10-15
u32 audioRate in openavb_avdecc_read_ini_pub.h, Notice about the 6188…

chris-kuhr:
…3 static values, need to be filled in with configuration parameters, and set dbs equal to mbla count in openavb_descriptor_stream_io.c


andrew-elder:
@chris-kuhr - thank you for this update. This looks good to me. Anyone else concur?

andrew-elder:
Thanks @bdthomsen! Merging.

PR #913, opened 2021-05-24
Added launchtime flag to talker

MarcinMiklas:
Sometime ago spin wait was made separate from fixed_timestamp, this caused regression in builds using IGB_LAUNCHTIME_ENABLED=1.

This PR adds *launchtime* flag to talker which will be set automatically if IGB/ATL_LAUNCHTIME_ENABLED is used.

Otherwise we would need to use spin_wait flag to make launchtime work which would be counter-intuitive.

Both spin_wait and launchtime mode work on WALLCLOCK, but launchtime doesn't need any waiting because it will be done in ethernet driver.

So, we will have 3 modes for packet scheduling:
- sleep
- spin_wait
- launchtime

### DIFF
--- a/lib/avtp_pipeline/tl/openavb_talker.c
+++ b/lib/avtp_pipeline/tl/openavb_talker.c
@@ -127,7 +127,7 @@ bool talkerStartStream(tl_state_t *pTLState)
 	// setup the initial times
 	U64 nowNS;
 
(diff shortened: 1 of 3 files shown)


andrew-elder:
This looks great! Any other comments?

bdthomsen:
I approve.

PR #915, opened 2021-05-24
Configuration options can be changed at runtime

MarcinMiklas:
This will allow creating GUI for openavb_harness. 

Example use cases.

To change volume at runtime enter in openavb_harness:

    c0 intf_nv_volume -20

To change filename of played wav file:

    c1 intf_nv_file_name test2.wav

c0, c1 means stream 0, stream 1 respectively.

Of course, it is not particularly handy to enter whose commands manually, but you can write the GUI wrapper which runs the openavb_harness and then issues those commands in response to user actions.


andrew-elder:
This looks ok to me. Any other comments?

bdthomsen:
It looks OK to me as well, though I would have appreciated some comments describing what the new functions do.

bdthomsen (approved):
Looks good to me.  Thanks for the updates.

PR #916, opened 2021-05-24
Api for packet dropping

MarcinMiklas:


andrew-elder:
Looks good to me. Any other comments?

lib/avtp_pipeline/avtp/openavb_avtp.c
```diff
@@ -384,11 +384,16 @@ openavbRC openavbAvtpTx(void *pv, bool bSend, bool txBlockingInIntf)
...
 			// Increment the sequence number now that we are sure this is a good packet.
 			pStream->avtp_sequence_num++;
-			// Mark the frame "ready to send".
-			openavbRawsockTxFrameReady(pStream->rawsock, pStream->pBuf, avtpFrameLen + pStream->ethHdrLen, timeNsec);
-			// Send if requested
-			if (bSend)
-				openavbRawsockSend(pStream->rawsock);
+
+			if (pStream->drop_avtp_packets) {
+				pStream->drop_avtp_packets -= 1;
```
bdthomsen, line 389: I'd prefer "if (pStream->drop_avtp_packets > 0)" to ensure we don't drop packets if drop_avtp_packets becomes negative.

MarcinMiklas: Maybe I should just make drop_avtp_packets type unsigned?

bdthomsen: Unsigned could be worse, as you will get a very large number if subtracting 1 from 0, so you would need to check for an upper bound (as well as not zero).  I suggest sticking with signed int.

bdthomsen (approved):
Looks good to me.  Thanks!

PR #917, opened 2021-05-24
Map AAF uses userspecific format for little endian

MarcinMiklas:
OpenAvnu is great test tool for other AVB stacks. Sometimes on MCUs we have to use non-standard AAF format for performance reasons. Like when little-endian MCU have to handle 32 channels of 32 bit class A low latency audio. Byte swapping in software is just too much.

This fix will allow talkers to use little-endian settings.


andrew-elder:
Minor change - I'm merging it.

